PR #9, opened 2019-06-16
Make the scssphp website more mobile-friendly

allejo:
- For GH pages, the `Gemfile.lock` is ignored so there's no need to track it
- Made some minor changes to the Sass to make the site more mobile-friendly
- Not sure why jQuery was being loaded but I don't see any usage

Preview: http://projects.allejo.io/scssphp/


robocoder:
IIRC jquery was used on the change log to show/hide logs. I don't think I ported that feature when migrating to jekyll

PR #23, opened 2019-07-24
Implemented passing Arguments to Content Blocks

jensniedling:
Due to personal requirement I implemented the option to pass variables into the content block provided to mixins. See the official documentation about this feature for details:

https://sass-lang.com/documentation/at-rules/mixin#passing-arguments-to-content-blocks

Since I'm really new to this project the will most likely be less well done parts about this implementation, that should be improved when adding it to the project. Especially the following two places:
- Compiler.php Line 2520 / Passing the argument list with a special variable similar to the content part. (Can this be avoided?)
- Compiler.php Line 2553 and 2572 / Extracting the required variables (I assume there's a simpler way I did not find) and applying those to the contents scope.

Unfortunally I am lacking the insight into the library to figure out a better solution or even tell if there is one. (especially the exact relations for env / storeEnv / scope part, which is really hard to get behind)

Feel free to optimize those parts, or help me understand the possible better solutions and I will adjust and make another pull request afterwards.

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -2467,7 +2467,7 @@ class Compiler
 
             case Type::T_INCLUDE:
                 // including a mixin
-                list(, $name, $argValues, $content) = $child;
+                list(, $name, $argValues, $content, $argUsing) = $child;
 
                 $mixin = $this->get(static::$namespaces['mixin'] . $name, false);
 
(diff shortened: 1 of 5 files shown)


Cerdic:
At first sight it looks good! I will checkout localy and do some checking before merging.

I can confirm that the env/storeEnv/scope is really a hard thing, and I'm still not fully able to figure the whole thing :)

Cerdic:
I cherry-picked the commit https://github.com/scssphp/scssphp/commit/84c922c8eed2bc6606de05dcc7be634086b8ab6b
and added some refactoring around the env/applyArgs to be able to have only one path for compiling content whether there is using args or not https://github.com/scssphp/scssphp/commit/e37b8751fc38b54586497910c857211fe2c6e0af
I think this is all good 
Thanks!

Issue #29, opened 2019-08-22
@each is forcing unwanted type conversion


devdot:
The types of keys are converted in a `@each` loop. 

Example (index not found for key in `@each` loop):
````SCSS
$map1: (
  "key1": 23,
  "key2": 32
);
$map2: ("key1", "key2");
@each $key, $value in $map1 {
  @debug $key; // will output as keyword, not as string
  @debug index($map2, $key); // will be null, but should be a number 
}
````

This was probably forgotten in #3 and is a simple fix.

Also, this fix makes it possible to compile Bootstrap 4 bootstrap-grid.scss without errors (I'm still having errors with bootstrap.scss but that's another issue)

robocoder:
Fixed in 01390f285f72cbe34f431c15aa7f2bc9025efc75

PR #30, opened 2019-08-22
Fix: @each is forcing unwanted type conversion #29

devdot:
I don't get the test to run so I can't add a unit tests at this point. Fix works as intended. See issue #29


robocoder:
Can you add a unit test?

robocoder:
Thanks!

PR #33, opened 2019-08-24
Fix zip function to allow all types as arguments #32

### LINKED ISSUE #32: Zip function not working for non-list types
This bug prevents Bootstrap 4.3.1 from compiling.

This kind of usage of `zip` will fail (expected list). However, it works in libscss and is used by Bootstrap.
````SCSS
zip(inline, 1px);
````

No sure if there is any reason to disallow this usage

devdot:
Fixes #32

Compiles latest version of Bootstrap: https://github.com/twbs/bootstrap/commit/b02bae769e989838229c13a7c97d5af4c338601d

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -6318,7 +6318,13 @@ class Compiler
 
     protected function libZip($args)
     {
-        foreach ($args as $arg) {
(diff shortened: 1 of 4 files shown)


robocoder:
Thank you

PR #42, opened 2019-09-10
Clean the phpunit setup

stof:
- Migrate to namespaced classes directly
- Ignore the `.phpunit.result.cache` file generated by recent versions of PHPUnit during a test run so that it does not get committed by mistake
- Improve the phpunit requirement


robocoder:
Thanks!

PR #43, opened 2019-09-10
Remove the PHP nightly job on Travis

stof:
nightly on Travis installs PHP 8.0-dev, but there is no PHPUnit version supporting it yet.


robocoder:
ok

PR #57, opened 2019-11-12
Fix importing of css files

gilbertsoft:
According to https://sass-lang.com/documentation/at-rules/import#importing-css a plain old css file can be imported providing the file name without extension. This patch fixes this feature.

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -4435,6 +4435,7 @@ class Compiler
                     $full = $dir . $separator . $full;
 
                     if (is_file($file = $full . '.scss') ||
+                        is_file($file = $full . '.css') ||
                         ($hasExtension && is_file($file = $full))
                     ) {
                         return $file;


Cerdic:
The feature should also include
> CSS files imported by Sass don’t allow any special Sass features. In order to make sure authors don’t accidentally write Sass in their CSS, all Sass features that aren’t also valid CSS will produce errors

Whereas
> LibSass supports importing files with the extension .css, but contrary to the specification they’re treated as SCSS files rather than being parsed as CSS. This behavior has been deprecated, and an update is in the works to support the behavior described below.

Not sure if it's better to move to a bad dereprecated partial support than no support at all as now.

A compromise could be to include as static content, whitch at least would avoid any SCSS compilation, but would come without the CSS syntax checking and the ability to `@extend`

gilbertsoft:
Yeah, saw that too. Will have a look if I can fully support the feature as mentioned in the docs....

gilbertsoft:
After having a deeper look I don't find an easy solution to achieve this. Imho a way could be to add an other parameter to compileChildrenNoReturn to flag plain css files and add a white list for allowed compilation scenarios as `extend`, everything else will throw an exception. But this needs a complete rework of every compile method and I don't know if this effort is useful as the whole @import stuff will be replaced by @use at some point.

So I'm going to close this PR and go a head with an other solution for the moment.

PR #63, opened 2019-11-29
added the variables to the serve method of the Server class

iannacone:


iannacone:
wrong, sorry.
i misunderstood the meaning of serve.
actually the vars should be included before compiling, right?
can it be done without that i pull a request?

PR #81, opened 2020-02-17
Init $libName before using it

enricobono:
We have a bunch of custom functions, registered via the `$compiler->registerFunction()`.

When scssphp calls the `->callNativeFunction()`, the code gets the custom function, but it fails some line afterwords, because it tries to use the var `$libName`, which is not initialised if the function is a custom one.
The code fails with this error:
> Undefined variable: libName

As you can see from the code, if you enter the first `if`, the var `$libName` never gets initialized.

This PR fixes this wrong behaviour.

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -4811,6 +4811,7 @@ class Compiler
         if (isset($this->userFunctions[$name])) {
             // see if we can find a user function
             list($f, $prototype) = $this->userFunctions[$name];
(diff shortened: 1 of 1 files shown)


robocoder:
Thanks.

I think this PR introduces a naming convention for user functions that wasn't intended. Just going to quick-fix this in the meantime.

enricobono:
Yup, agree.
Thank you for the fix.

enricobono:
@robocoder Could you please bump the version number, so we can actually use the fixed version of the library? 😅 
Thanks in advance 😊

PR #98, opened 2020-04-17
Qualify function calls when the compiler can optimize them

### LINKED ISSUE #96: Null checks are a performance bottleneck of this library
<details>
<summary>Some context about how I went there and what the profiles are about</summary>

At Incenteev, I have a use-case where I need to render some scss code at runtime (based on some values configurable by admin users). I first implemented this with scssphp (well, actually, the first one used lessphp as we migrated from less to scss). Due to the slowness of scssphp, I once decided to implemented an alternative way, which would use symfony/process to call a node.js script which would compile things and return the output. That was worth it, as even with the overhead of launching a separate process, things got significantly faster. For some reasons, I had to keep the PHP-based implementation available even though the prod was using the node-based renderer. And while refactoring the node code to use `sass` rather than `node-sass` (loosing some performance there, but in favor of other benefits), I decided to compare the performance of the PHP-based renderer again.
The blackflire profiles I'm sharing are about using this endpoint, running on my local machine, with Symfony in the prod environment and an optimized composer autoloader.

<small>Note that my controller actually involves some caching to return the cached CSS in a few ms when the config has not change. I disabled this caching layer while profiling</small>
</details>

While profiling the code rendering scss with Scssphp 1.0.8, here is what appeared in blackfire:

image

Having `is_null` as the second most expansive function looked surprising to me. Among these 3.6 millions calls, more than 2.8 million calls correspond to a single place in the source code (inside a loop obviously): the check in `Compiler::callStackMessage`:

image

I tried a simple 1-char patch in this method:

```diff
     protected function callStackMessage($all = false, $limit = null)
     {
         $callStackMsg = [];
         $ncall = 0;

         if ($this->callStack) {
             foreach (array_reverse($this->callStack) as $call) {
                 if ($all || (isset($call['n']) && $call['n'])) {
                     $msg = "#" . $ncall++ . " " . $call['n'] . " ";
                     $msg .= (isset($this->sourceNames[$call[Parser::SOURCE_INDEX]])
                           ? $this->sourceNames[$call[Parser::SOURCE_INDEX]]
                           : '(unknown file)');
                     $msg .= " on line " . $call[Parser::SOURCE_LINE];

                     $callStackMsg[] = $msg;

-                    if (! is_null($limit) && $ncall > $limit) {
+                    if (! \is_null($limit) && $ncall > $limit) {
                         break;
                     }
                 }
             }
         }

         return implode("\n", $callStackMsg);
     }
```

And here is the result of this patch: 
image

The reason is that null-checks can be optimized by the PHP compiler when it knows about them. There is 3 ways to do this check in namespaced code:

- a strict comparison (`$limit === null`)
- a qualified function call (`\is_null($limit)`)
- a non-qualified function call, relying on the global function fallback (`is_null($limit)`).

The compiler cannot resolve function fallbacks (that must happen at runtime, after checking whether a `Scssphp\Scssphp\is_null` function exists), and so we end up with a function call rather than a very optimized opcode.

Given the massive optimization opportunity (bigger than this 28% as I changed a single place in the code, and other null checks still make `is_null` be ranked 6th for now. It goes to 31% when qualifying all `is_null` in the Compiler class), I suggest changing the way null-checks are performed in the codebase.

There is 2 solutions, and I will let the maintainers make a choice. **Both solutions can be automated with PHP-CS-Fixer**.

- use comparisons (use the `is_null` fixer)
- qualify the function calls which are optimized by PHP (use the `native_function_invocation` with the `@compiler_optimized` set of functions included)

> Note that both fixers are considered risky. That's because they will remove the fallback behavior and so will break code when you actually have a namespaced `is_null` function. But that's not an issue in this project.

Note that whatever decision you make for the way to check for `null` in the project, I suggest adding the `native_function_invocation` fixer with `@compiler_optimized` in the config, because `is_null` is not the only function being optimized.

stof:
Some built-in functions can be turned into specific opcodes at compile time when the compiler is sure that they will be used (i.e. no namespace fallback is needed). This can have a major performance impact when they appear in hot paths.
In the profiling, is_null was one of the most expansive function due to the huge number of times it was used. And it is one of the functions which is optimized. Closes #96.

This change has been done using the php-cs-fixer tool, which maintains a list of functions which benefit from such optimization.

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -102,13 +102,13 @@ class Cache
         ) {
             $cacheTime = filemtime($fileCache);
 
-            if ((is_null($lastModified) || $cacheTime > $lastModified) &&
+            if ((\is_null($lastModified) || $cacheTime > $lastModified) &&
                 $cacheTime + self::$gcLifetime > time()
             ) {
                 $c = file_get_contents($fileCache);
                 $c = unserialize($c);
 
-                if (is_array($c) && isset($c['value'])) {
+                if (\is_array($c) && isset($c['value'])) {
                     return $c['value'];
                 }
             }
--- a/src/Colors.php
+++ b/src/Colors.php
@@ -186,7 +186,7 @@ class Colors
      */
     public static function colorNameToRGBa($colorName)
     {
-        if (is_string($colorName) && isset(static::$cssColors[$colorName])) {
+        if (\is_string($colorName) && isset(static::$cssColors[$colorName])) {
             $rgba = explode(',', static::$cssColors[$colorName]);
 
             // only case with opacity is transparent, with opacity=0, so we can intval on opacity also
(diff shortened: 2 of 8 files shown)


stof:
Here is the command I ran to generate this diff:

```
php-cs-fixer fix src --rules='{"native_function_invocation": {"scope":"namespaced","include":["@compiler_optimized"]}}' --allow-risky=yes
```

Issue #89, opened 2020-03-02
Infinit loop with combination Foundation Zurb 6


havutcuoglu:
Hi there,

in combination with Foundation ZURB 6 I get infinite loop problems when I use `@include breakpoint (...)`. Sometimes the infinite loop happens because of the `rem-calc()` function.

I import the following lines into my app.scss:
```
@import "foundation-sites / scss / foundation";
@include foundation-everything;
// --------
@import "mixins";
```
And I only have the following lines in the mixin file:
```
// Spacing as Margins / Paddings
// - - - - - - - - - - - - - - - - - - - - - - - - - - -

$length: 20;
$factor : 1rem;

@mixin spacing-classes {
    @for $i from 0 through $length {
        $value: $i * $factor;

        // margin classes
        .m-#{$i} {margin: $value;}
        .mt-#{$i} {margin-top: $value;}
        .mr-#{$i} {margin-right: $value;}
        .mb-#{$i} {margin-bottom: $value;}
        .ml-#{$i} {margin-left: $value;}
        .mx-#{$i} {margin-left: $value; margin-right: $value;}
        .my-#{$i} {margin-top: $value; margin-bottom: $value;}
        // padding classes
        .p-#{$i} {padding: $value;}
        .pt-#{$i} {padding-top: $value;}
        .pr-#{$i} {padding-right: $value;}
        .pb-#{$i} {padding-bottom: $value;}
        .pl-#{$i} {padding-left: $value;}
        .px-#{$i} {padding-left: $value; padding-right: $value;}
        .py-#{$i} {padding-top: $value; padding-bottom: $value;}
    }
    @include breakpoint(small only) {
        @for $i from 0 through $length {
            $value: $i * $factor;
            // margin classes for small screens
            .s-m-#{$i}  {margin: $value;}
            .s-mt-#{$i} {margin-top: $value;}
            .s-mr-#{$i} {margin-right: $value;}
            .s-mb-#{$i} {margin-bottom: $value;}
            .s-ml-#{$i} {margin-left: $value;}
            .s-mx-#{$i} {margin-left: $value; margin-right: $value;}
            .s-my-#{$i} {margin-top: $value; margin-bottom: $value;}
            // padding classes for small screens
            .s-p-#{$i} {padding: $value;}
            .s-pt-#{$i} {padding-top: $value;}
            .s-pr-#{$i} {padding-right: $value;}
            .s-pb-#{$i} {padding-bottom: $value;}
[...]

PR #99, opened 2020-04-17
Fix leaks in the call stack

stof:
When using a return directive, a frame was leaked in the call stack.
This has 2 effects:

- the call stack displayed in error message becomes very confusing, as it includes unrelated locations
- the call stack grows much more than necessary. This has a cost when computing the call stack message, and makes it more likely to reach the threshold used to prevent infinite loops (without actually doing infinite loops).

Another frame leak happened when using a return directive in a forbidden location (so now leaking 2 frames due to the previous leak), in case exceptions are ignored (to allow the code to continue to execute and so to care about it).

This will probably solve #89 but I haven't checked.

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -1891,6 +1891,7 @@ class Compiler
             $ret = $this->compileChild($stm, $out);
 
             if (isset($ret)) {
+                $this->popCallStack();
                 return $ret;
             }
         }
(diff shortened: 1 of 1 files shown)


havutcuoglu:
Hey @stof it's work fine.

I change the call stack again on 25,000 and make your changes from your [commit](https://github.com/scssphp/scssphp/pull/99/commits/c3a80d7aa148c4a40483c68ea41856f2f113d81e) and it's work.

Thank you vey much.

PR #100, opened 2020-04-17
Add named call stack entries for imports

stof:
This makes the call stack much more helpful, as it gives us the chain of imports leading to the error, not only the chain of function calls. This matches the behavior of dart-sass and libsass regarding error reporting.

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -4607,6 +4607,7 @@ class Compiler
      */
     protected function importFile($path, OutputBlock $out)
     {
+        $this->pushCallStack('import '.$path);
         // see if tree is cached
         $realPath = realpath($path);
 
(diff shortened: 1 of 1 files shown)


stof:
I don't know how to add test coverage for that. AFAICT, the only tests covering the error reporting are in ExceptionTest, which does not rely on file-based fixtures, and so cannot easily have imports.

stof:
Here is a before/after comparison, when applying this patch on top of #99 (to avoid having a mess due to the leak).

Before:

```
Mixin or function doesn't have an argument named $b.: /home/stof/src/Incenteev/Incenteev/assets/scss/app/themes/../_common_theme.scss on line 14, at column 1
```

After:

```
Mixin or function doesn't have an argument named $b.: /home/stof/src/Incenteev/Incenteev/assets/scss/app/themes/../_common_theme.scss on line 14, at column 1
Call Stack:
#0 import /home/stof/src/Incenteev/Incenteev/assets/scss/app/themes/../_common_theme.scss /home/stof/src/Incenteev/Incenteev/assets/scss/app/themes/../_simple_theme.scss on line 19
#1 import /home/stof/src/Incenteev/Incenteev/assets/scss/app/themes/../_simple_theme.scss (unknown file) on line 4
```

PR #115, opened 2020-06-14
add argument to user functions callback

ryssbowh:
Hi,

I needed to extract the assets defined in node modules scss files (images, fonts), I figured the way to do it was to define the url function in the compiler, but the callback didn't tell me which file was being treated so I added it to the callback. Sharing it in case you think it's useful.

Thanks

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -4958,7 +4958,7 @@ class Compiler
             }
         }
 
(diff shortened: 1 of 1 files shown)


robocoder:
Added Compiler `getSourcePosition()` instead.

PR #117, opened 2020-06-26
Add access to scss exception line.

cbl:
I need the ability to use the `scss` exception line to display more readable errors directly in the template. For this I need the line where error happens. If the scss exception line is passed to the `ParserException` you can get it directly without reading it from the error message.

```php
try {
    return (new ScssPhp())->compile($style);
} catch (ParserException $e) {
    $e->getStyleLine();
}
```

### DIFF
--- a/src/Exception/ParserException.php
+++ b/src/Exception/ParserException.php
@@ -11,6 +11,8 @@
 
 namespace ScssPhp\ScssPhp\Exception;
 
+use Exception;
+
 /**
  * Parser Exception
  *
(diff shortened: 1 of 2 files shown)


robocoder:
Added source position getter/setter to ParserException

cbl:
Thanks a lot.

Here's the commit for anyone searching for the commit: https://github.com/scssphp/scssphp/commit/d12a6ba0ff82ea2ad187fc854641845e0bd73ee4

Issue #120, opened 2020-07-01
Undefined offset: 2 in Compiler.php (line 5209)


chrisblakley:
When compiling today on one of my websites, I received this notice:

`Undefined offset: 2 in /scssphp/src/Compiler.php:5209`

Here is the stack trace that I could capture for reference:
<img width="967" alt="Screen Shot 2020-07-01 at 8 29 47 AM" src="https://user-images.githubusercontent.com/6349678/86244408-ed002980-bb75-11ea-9444-641ff0642a95.png">

It seems that `$arg[2]` may not always exist for this condition in that file:
`} elseif ($arg[2] === true) {`

I wish I had more information to recreate this, but I am making a pull request to coincide with this issue that changes it to the following (which matches the above conditional for `$arg[0]`):
`} elseif (! empty($arg[2])) {`

Cerdic:
Indeed, the third arg is set here
https://github.com/scssphp/scssphp/blob/master/src/Parser.php#L2340
but sometimes not
https://github.com/scssphp/scssphp/blob/master/src/Parser.php#L2250

Cerdic:
This is solved by cb1c5c612e949aa2c9359be7dbcb3c25e87fa4e3

PR #121, opened 2020-07-01
Prevent "Undefined offset: 2" notice

chrisblakley:
Addresses #120 where `$arg[2]` may not exist


Cerdic:
I cherry-picked the commit, thanks!

PR #126, opened 2020-07-23
[TASK] Set mkdir recursive mode for creating subfolders in cache path

mobasoft:
I had to set mkdir recursive mode for creating subfolders

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -177,7 +177,7 @@ class Cache
         self::$cacheDir = rtrim(self::$cacheDir, '/') . '/';
 
         if (! is_dir(self::$cacheDir)) {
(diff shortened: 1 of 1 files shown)


robocoder:
0777 is considered a security issue

As a library, I'm inclined to remove the mkdir altogether and to disable the cache if the cacheDir is not writeable.

mahagr:
@robocoder Except it isn't. If you read the documentation, it reads that the method is using `umask()`, which prevents the directory from being created with widest permissions. And BTW, the code uses the default value from the method. :)

mahagr:
Symfony also defaults to 0777, and so does almost every library out there:

https://github.com/symfony/filesystem/blob/master/Filesystem.php#L89-L96

PR #127, opened 2020-08-17
Load sass-spec from the official repository

### LINKED ISSUE #102: Sass-spec should be updated to use the official source
According to the readme, the sass-spec being used for the testsuite is taken from the libsass repo.
The official source is at https://github.com/sass/sass-spec so I think it would make sense to take them from there instead (and maybe reference them as a git submodule or install them with a composer inline package repository rather than copying them in the repo).

stof:
This makes it clear which version of the spec is being tested, and makes sure that no hrx file gets forgotten in the import (which was the case).

Closes #102


stof:
Changes in tests/specs/sass-spec-exclude.txt are because one of the hrx file was missing in the imported specs, so this changes the total.

Cerdic:
Unfortunately, on MacOS this is not working (at least on my computer?) because of encoding/caps on some test file
```
composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 1 install, 0 updates, 0 removals
  - Installing sass/sass-spec (3.6.0): Loading from cache
    Failed to execute unzip -qq  '/Users/cedric/Sites/libs/scssphp/vendor/sass/sass-spec/10a49f981ddfd1d727953d635c1f9f2b' -d '/Users/cedric/Sites/libs/scssphp/vendor/composer/734297d3'

error:  cannot create /Users/cedric/Sites/libs/scssphp/vendor/composer/734297d3/sass-sass-spec-eb1d9a1/spec/libsass/Sa��ss-UT��F8.hrx
        Illegal byte sequence
error:  cannot create /Users/cedric/Sites/libs/scssphp/vendor/composer/734297d3/sass-sass-spec-eb1d9a1/spec/libsass/S+�ss-U+�F8.hrx
        Illegal byte sequence

    The archive may contain identical file names with different capitalization (which fails on case insensitive filesystems)
    Unzip with unzip command failed, falling back to ZipArchive class
Package phpunit/php-token-stream is abandoned, you should avoid using it. No replacement was suggested.
Generating autoload files
```

The result is I indeed have one file less on my hard disk and the exclude file is misinterpreted, leading to a complete fail of the tests :(

Cerdic:
This is because of the 2 files
https://github.com/sass/sass-spec/blob/master/spec/libsass/Sa%CC%81ss-UT%CC%A7F8.hrx
https://github.com/sass/sass-spec/blob/master/spec/libsass/S%C3%A1ss-U%C5%A2F8.hrx

I can go around this problem by making a manual copy of one of this file onto the name `Sass-spec-UTF8` then I have the right number of tests on my hard disk and everything is fine, but this is not super friendly for a developper that want to enters the project and also use MacOs (or other case insensitive File System)

stof:
So, the sass-spec repo cannot be installed on macOS ? What happens if you try to clone their repository ?

Cerdic:
On a git clone I have the same issue:
```
$ git clone https://github.com/sass/sass-spec.git
Cloning into 'sass-spec'...
remote: Enumerating objects: 65, done.
remote: Counting objects: 100% (65/65), done.
remote: Compressing objects: 100% (53/53), done.
remote: Total 58049 (delta 30), reused 27 (delta 10), pack-reused 57984
Receiving objects: 100% (58049/58049), 7.39 MiB | 3.55 MiB/s, done.
Resolving deltas: 100% (21180/21180), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'spec/libsass/Sáss-UŢF8.hrx'
  'spec/libsass/Sáss-UŢF8.hrx'
```

You can clone it, but only one of the two files is in the repo

stof:
I suggest opening an issue upstream then, and I will put this work on hold for now (basing my other PR on master directly)

Cerdic:
At least, the test list is now not sensitive to the test order/total number
https://github.com/scssphp/scssphp/commit/d5bfd57575e277b97975284f64dcf8500ab296f0

stof:
good

stof:
@Cerdic the files are still built with the numbers in them, which does not make them non-sensitive to conflicts when rebuilding them.

stof:
@Cerdic can you explain me how these numbers are meant to help debugging ? In the number `1959/1980.`, `1959` is *not* the value to pass to `phpunit --filter` to run only that test, because the dataset name is 1958 (keys are 0-based indices in the data-provider return value while these names are 1-based).
I think we should rather remove them entirely

Cerdic:
@stof yes you are right, il will remove the numbers from the exclude list.
In fact I'm using the display of failing test, during the test itself to get the number failing 
(and a quick and dirty script that extract the matching test case in a scss file to run it separatly and allow free debug)

In addition, removing all the numbers allow to integrate this PR, as the missing of one test on macOs when cloning should not have any other impact then

Cerdic:
This is integrated then with 624c78ea d6a149e6 and ac55cd77

(The MacOS issue is no more a real issue and we can safely ignore it)

PR #128, opened 2020-08-17
Properly skip error tests for sass-spec in compat run

stof:
Instead of doing a test without any assertion (which gets reported as risky), this reports the test as skipped, which matches the actual behavior.
Without `TEST_SASS_SPEC=1`, these tests are always skipped by the exclusion list anyway.


Cerdic:
Thanks, I cherry-picked on https://github.com/scssphp/scssphp/commit/7f5cd1f1c09267c7837b2b7a54e1f1e98168d399

PR #129, opened 2020-08-17
Refactor the handling of warnings in SassSpecTests

stof:
Instead of prepending them to the output and expectation to do a single assertion, the assertion is now done separately. This makes it much less confusing when the ouput is the right one but the warning is missing.
A new exclude list file is maintained, to cover tests for which the compiler generates the right output but does not display the warning, instead of skipping them entirely. This allows for more tests to run against the spec (missing a warning is less an issue than getting the output wrong). And this matches the behavior of libsass (the options.yml in hrx files can declare warnings as not implemented by libsass).

Note that for now, I don't try to read the options.yml to know whether libsass is missing the warning or no. the warning exclusion list is entirely maintained in this repo, as the missing warnings might not be the same than libsass

~~Currently includes #127 to avoid conflicts on the exclude list due to the additional hrx file included in it (this is why this PR is a draft)~~ not anymore as the other one is not ready yet


stof:
@Cerdic this one is now ready for review

Cerdic:
Cherry-picked in https://github.com/scssphp/scssphp/commit/7aa207f47d73369b87c3d591c63449758c34d64a

PR #130, opened 2020-08-17
Ensure sass-spec tests are properly isolated for precision changes

stof:
`Compiler::setNumberPrecision` is actually changing some global state. For now, sass-spec tests are not changing this config yet (which is why some specs are skipped) so this is not an issue. But it breaks test isolation if we want to start testing with different precision.

### DIFF
--- a/tests/SassSpecTest.php
+++ b/tests/SassSpecTest.php
@@ -13,6 +13,7 @@ namespace ScssPhp\ScssPhp\Tests;
 
 use PHPUnit\Framework\TestCase;
 use ScssPhp\ScssPhp\Compiler;
+use ScssPhp\ScssPhp\Node\Number;
 
 /**
  * Sass Spec Test - extracts tests from https://github.com/sass/sass-spec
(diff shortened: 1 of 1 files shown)


stof:
Note that a better solution might actually be to stop using a static state for that, which would actually allow using 2 separate Compiler instances with their own settings (be it in parallel or in sequence, btw, as the Compiler does not reset the global state on instantiation either).
Is it allowed to change the number precision after starting the parsing but before ending the compilation or can it never change at this point ? I think moving this state to a compiler instance value that gets copied into the Number node is easy to do as refactoring if you are only allowed to change the precision before calling `compile` (and so a Number node does not care if the compiler precision changes after it got instantiated).

Cerdic:
cherry-picked in acec6296

stof:
@Cerdic what about my comment about avoiding that global state ?

Cerdic:
I see at least one case when the number precision can change between the parsing and the compilation : it's because the parsing result can be cached and thus the parsing and the compilation can occurs at different moments.

But as I can see the number precision is only used in the `output()` method that also recieve the `$compiler` instance as an argument, so the precision could be a property of the Compiler, the `output()` using it if possible and otherwise using a `static::$defaultPrecision` (this is when called by the `__toString()` method)

stof:
Irrelevant anyway, as the precision is not configurable anymore at all now.

PR #131, opened 2020-08-17
Respect the compiler precision when computing the alpha channel

stof:
This fixes 2 specs


Cerdic:
cherry-picked in 00a26dc8

PR #132, opened 2020-08-17
Fix the parsing of id tokens in values

stof:
If the token was starting with valid hex chars, the parser was previously splitting it into 2 keywords, with a color for the beginning.


Cerdic:
cherry-picked in 29285f95

PR #137, opened 2020-08-17
fix the sass-spec tests when the warning is different

stof:
When refactoring the way warnings are asserted in the testsuite (in #129), a mistake was done, making tests fail when Scssphp was triggering a warning. Fixing this makes more tests go from being excluded entirely to excluding only the warning assertion (as the formatting of the warning is different).


Cerdic:
cherry-picked in e5f661cb

PR #138, opened 2020-08-18
Refactor the format of the output for debug and warn directives 

stof:
The output format of `@debug` and `@warn` now matches libsass, which unlocks some tests


Cerdic:
Sorry, I saw too late this was still a draft branch and I cherry-picked it before refactoring the exclude list to avoid any conflict when getting it! Thanks for that and all the others PR :)

(cherry-picked in 0cf18444 and e5f661cb)

stof:
The draft state was only because of the first commit being #137

PR #139, opened 2020-08-18
Simplify the handling of comments to be more spec-compliant

stof:
The special logic dealing with comments was actually making them less spec-compliant than without it, as sass-spec differs from the old ruby-sass logic.

The output is a bit less nice for multi-line comments as it does not fix their indentation, but that matches the libsass behavior.
dart-sass has a slightly different output, where the re-indent the comment by replacing the original indentation with the output indentation. Something like this could be achieved in the nested and expanded formatters too. I haven't done it yet, as it would make the sass-spec compliance harder to maintain (the testsuite currently uses the libsass output when the hrx file contains implementation-specific outputs).

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -2381,14 +2381,6 @@ class Compiler
     {
         $outWrite = &$out;
 
-        if ($type === Type::T_COMMENT) {
-            $parent = $out->parent;
-
-            if (end($parent->children) !== $out) {
-                $outWrite = &$parent->children[\count($parent->children) - 1];
-            }
-        }
-
         // check if it's a flat output or not
         if (\count($out->children)) {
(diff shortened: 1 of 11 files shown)


Cerdic:
I just sent a modification to accept any test that match one of the provided ouput and not only the libsass output, because we don't need to stick to the libsass output as long as the output is acceptable according to the spec

81cf931d2d803e2b246b953d78684d268ef99a42

Cerdic:
cherry-picked in bc6cd5d8

stof:
> I just sent a modification to accept any test that match one of the provided ouput and not only the libsass output, because we don't need to stick to the libsass output as long as the output is acceptable according to the spec

that's great

PR #140, opened 2020-08-18
Reorder color names to ensure that the same name is used than in libsass

stof:
Several colors have multiple names in CSS (including all gray colors). The color map now ensures that the first name in the array is used, and that colors are ordered in the same way than in libsass so that the final name is the same.
Note that dart-sass uses a different rule for the canonical name, always using the first one in alphabetical order (which matches libsass for gray/grey but not for magenta/fuschia or cyan/aqua). I went for the libsass rules because that's how sass-spec is integrated right now.


stof:
I just saw that you made the dart-sass output acceptable, so I will update this PR with my previous list instead. I think using the reference implementation for the color names is better.

Cerdic:
cherry-picked in 03372d70

stof:
hmm, too late :smile:

PR #142, opened 2020-08-18
Improve test names for sass-spec by using a named provider

stof:
This makes it a lot easier to use phpunit --filter as using the file name (without the hrx) exception now works rather than having to
discover the index.
The number of tests is back to 1-based in the names to make it less confusion (tests now go again from 1/1980 to 1980 rather than from 0/1980 to 1979/1980), which is not an issue as phpunit --filter does not require the usage of a 0-based index anymore.

### DIFF
--- a/tests/SassSpecTest.php
+++ b/tests/SassSpecTest.php
@@ -453,12 +453,15 @@ class SassSpecTest extends TestCase
         ksort($tests);
         $tests = array_values($tests);
 
+        $testCases = array();
+
         foreach ($tests as $k => $test) {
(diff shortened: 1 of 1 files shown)


stof:
This has another benefit. I actually started that work to be able to identify which test was triggering an infinite loop when trying to run error tests from the specs. With readable names, I could run it with `phpunit --textdox` which outputs test names during the execution, and so find out the name of the test just before the looping one. With 0-based indices as test names, the `phpunit --textdox` output is almost useless as it won't tell us which spec was executed.

Cerdic:
cherry-picked in c91fbf53

PR #144, opened 2020-08-18
Execute error tests from sass-spec 

stof:
For now, such tests are expected to throw a SassException to report the error, but don't assert the error message yet, as the format is not normalized with libsass and dart-sass.

In the first commits of the PR, I fixed some cases where the compiler would either enter into infinite loops or trigger PHP errors in some cases (happening in the newly executed specs)


tests/SassSpecTest.php
```diff
@@ -187,28 +188,30 @@ public function testTests($name, $input, $output)
...
+                    passthru("mkdir -p " . dirname($f));
                 }
 
-                static::$scss->setImportPaths([$basedir]);
+                file_put_contents($f, $c);
             }
 
+            static::$scss->setImportPaths([$basedir]);
+        } else {
+            static::$scss->setImportPaths([]);
```
stof, line 207: this resetting ensures that we reset the state of the shared compiler even if a previous test failed to reset it properly (due to an error for instance), which improves test isolation.

I'm actually considering instantiating a separate compiler per test instead to improve isolation, but that's out of the scope of this PR.

Cerdic: I used to have compilation memory issue when wanting to run tests suite over the complete Sass-spec and I was finaly able to do that only with a static instance of the compiler.

But could also be because there was a lot of errors&fails at this time (and memory leaks?), maybe now it could work, so go for a try if you want!

stof: I tried it, and PHPUnit reports exactly the same memory usage

Cerdic:
fully cherry-picked

stof:
@Cerdic why not using the github merge feature in rebase mode rather than cherry-picking manually ? this would mark the PR as merged and github (and let the merged commit reference the PR)

Cerdic:
yeah you're right, I should try, but I'm pulling and testing on my computer before and like here I added a commit to fix the error stream on the tests expecting errors and at the end it's easier to just push all the rebase and close manualy.

Is there still a merge commit if I merge in rebase mode?

stof:
> Is there still a merge commit if I merge in rebase mode?

no

PR #151, opened 2020-08-18
[WIP] Refactor the handling of strings and escape sequences

### LINKED ISSUE #149: The handling of escape sequences is not spec compliant
Currently, they are stored in the string representation directly rather than being parsed. This causes several issues:

- double escaping in some cases
- missing escaping in the output in some cases
- wrong interpolation in some cases
- wrong result for string-manipulating functions

stof:
Closes #149


stof:
As a prototype for the decoding of Unicode code points, I used `mb_chr` for now. But this API exists only in PHP 7.2+ with `mbstring` loaded. That's 2 requirements that the parser does not have today.
A solution could be to copy the implementation available in the Symfony polyfill: https://github.com/symfony/polyfill/blob/1c93f786da6e8309434b53d3944a7e00264854c1/src/Mbstring/Mbstring.php#L731-L748. That would avoid adding a dependency on a polyfill package.

Cerdic:
Let's go for a copy of the Symfony polyfill, indeed

Cerdic:
I worked on top of this PR and was able to fix the issues coming from this refactoring.
So 5a8b113d and 04948e20 are cherry-pick of this PR
and then follow f810ddcc 84b8a009 d301b16f 95ef07d8 11247dd9 which closes #149 I think

PR #153, opened 2020-08-18
Fix the logic setting the memory_limit in SassSpecTest

stof:
The new version of the code makes sure that 256M is the minimum limit, but will preserve any higher limit.
This makes it possible to use PCOV to collect code coverage while running tests, as it requires lots of memory.


stof:
The logic to parse the memory_limit is taken from the composer source code.

PR #155, opened 2020-08-19
Upgrade sass-spec to the latest version

### LINKED ISSUE #152: [RFC] use an uptodate version of sass-spec for the testsuite
Currently, we use the libsass-3.6.0 tag of the sass-spec repository, which is 330 commits behind master as of today.

I tried updated sass-spec to https://github.com/sass/sass-spec/commit/73222792c42a516d62e2e25c3f5d9e35f5567030 (which is currently the master branch). And SassSpecTest runs a lot more tests:

- before: Tests: 1981, Skipped: 728  -> 1253 executed specs (which are passing)
- after: Tests: 3971, Skipped: 1544  -> 2427 executed specs (which are passing)

As we already have the infrastructure in place to manage skipping specs that are not supported, I suggest that we run an uptodate version of sass-spec rather than a version from 1 year ago, to benefit from all the new tests added in the meantime.
what do you think ?

stof:
As we don't support modules yet and we are not working on them for now, related specs are always skipped even when running with `TEST_SASS_SPEC=1`. These specs are not listed in `exclude.txt` files either, making it easier to use these files to track spec compliance on non-modules features.

Closes #152


composer.json
```diff
@@ -42,16 +42,16 @@
             "type": "package",
             "package": {
                 "name": "sass/sass-spec",
-                "version": "3.6.0",
+                "version": "2020.08.10",
```
stof, line 45: As sass-spec does not tag releases anymore, considering it a living spec, I decided to use date-based versions here. We don't really care about what the version is, as long as it is the same in `require-dev`. But it is important that this version gets changed when we change the referenced commit, so that `composer update` knows that this is a different version. Using the date is the easy way here (and compatible with composer) and gives an info about how uptodate the sass-spec is.

Cerdic:
here we go!

PR #158, opened 2020-08-19
Stop skipping all specs having a todo on them

stof:
The fact that they have a TODO on them for either libsass or dart-sass is not enough to decide to drop them without trying to make them pass. And some of them actually pass in scssphp. The normal exclusion logic is used to handle them instead.

The only remaining dropped tests are:
- tests using an input.sass rather than an input.scss to tests specifically the indented syntax of Sass, as scssphp does not support it
- tests missing an expected output or error (this happens when there is only an error-dart-sass as we don't support alternative error and warning yet)
- mis-detected tests not having an input for hrx folders containing only a readme while specs are in subfolders of the hrx. This is caused by the fact that we don't actually parse hrx files as archives but we rely on the sass-spec convention of adding a delimiter comment instead.
- huge.hrx, because it is too big and would cause memory usage issues.

SassSpecTest now runs 2269 specs, with 939 skipped and 1330 passing, compared to 1981 specs with 728 skipped and 1253 passing before that change.


tests/SassSpecTest.php
```diff
@@ -172,6 +171,14 @@ public function testTests($name, $input, $output)
             return;
         }
 
+        if (strpos($name, 'libsass-closed-issues/issue_1801/import-cycle') ||
+            strpos($name, 'libsass-todo-issues/issue_1801/simple-import-loop') ||
+            strpos($name, 'libsass-todo-issues/issue_221262') ||
+            strpos($name, 'libsass-todo-issues/issue_221292')
```
stof, line 177: this is moved here, because `issue_221292` is not an error test.

Also, I mark them as skipped all the time rather than putting them in the exclusion list, as we don't actually run them. This is similar to how I handle module specs in #155

Cerdic:
I cherry-picked in 376fcc9c to manage the conflict on exclude file list (I regenerated the list in fact)

stof:
Conflicts with the sass-spec upgrade have been fixed

Cerdic:
Sorry, was doing it in parallel as well :)

stof:
hmm, I pushed it too late, and you made a mistake in your conflict resolution. We now have duplicate lines for the destructuring of input and output variables.

Cerdic:
shit, seen it. I send the correction

PR #160, opened 2020-08-19
Stop using includes to filter out module-based specs

stof:
Now that includes are shared for the whole hrx file, this filters out too many specs. So the module filtering is now done only for the input file. Specs which use modules in the includes but import them in the old way in the input file are now handled by the exclude file instead.

Test summary for `phpunit tests/SassSpecTest.php`, showing the executed passing tests:

- https://github.com/scssphp/scssphp/commit/6db990e3ef2cbbf17f7d5c65943106e55d1273c7: Tests: 5453, Assertions: 2685, Skipped: 2789.
- this PR: Tests: 5453, Assertions: 2733, Skipped: 2741.


stof:
Also, if I compare the skipped tests when running `TEST_SASS_SPEC=1 phpunit tests/SassSpecTest.php` (which can be skipped only because of being detected as using modules or because of being one of the few tests triggering infinite loops), this looks like that:

- a2f45d0 (before the change): Skipped: 984
- 6db990e (current master): Skipped: 1011
- this PR: Skipped: 941 (this is because the filtering based on includes for same-spec includes was finding 43 specs using modules only in includes)

Cerdic:
I improved the includes managing in 734861f1 to try to get only the needed includes for each test and avoid the excluding caused by an include from a sibling test.
Is it good enough for you?

I would really prefer that the exclude list only list tests that we want to be able to pass.

By the way, maybe it should be renamed `sass-spec-todo.txt` to be more clear, and then we *could* have a real exclude list with manualy cherry-picked test we decide we don't want to try to target (at least for now)?

stof:
We're back at `Skipped: 984`, so that's fine with me. I'm closing this PR.

Cerdic:
For the record, there should be some plateform specific glitch somewhere cause I'm having 994 and not 984:
```
$ git lg -1
* 70611315 - (HEAD -> master, origin/master, origin/HEAD) PSR2 (Cerdic 13 minutes ago)
$ make sass-spec
TEST_SASS_SPEC=1 vendor/bin/phpunit --colors tests 2>&1 | tee /tmp/sass-spec.log | tail -2
ERRORS!
Tests: 5852, Assertions: 4812, Errors: 104, Failures: 1694, Skipped: 994.
```

(but this is not really important right now)

stof:
@Cerdic my count is done on `phpunit tests/SassSpecTest.php` only. There might be skipped tests elsewhere in the full testsuite

PR #173, opened 2020-09-25
Use PHP methods to do IO rather than a shell command

stof:
This has several benefits:

- it makes it possible to run tests on Windows (I haven't tried it yet to see whether they are passing or no, but the usage of `passthru` with a Unix command made that entirely impossible before)
- it goes a bit faster (the testsuite run takes a few seconds less, thanks to avoiding the overhead of spawning a child process)

As an additional benefit, I refactored the cleaning to be performed in a phpunit teardown method (using `@after` rather than overriding `tearDown` to avoid issues with the signature change in PHPUnit 8). this ensures that the cleaning also happens when an exception is raised, either because of a failure or because of expecting an exception (the cleaning on line 326 was never reached for passing tests)

### DIFF
--- a/tests/SassSpecTest.php
+++ b/tests/SassSpecTest.php
@@ -29,6 +29,19 @@ class SassSpecTest extends TestCase
 
     protected static $fileExclusionList = __DIR__ . '/specs/sass-spec-exclude.txt';
     protected static $fileWarningExclusionList = __DIR__ . '/specs/sass-spec-exclude-warning.txt';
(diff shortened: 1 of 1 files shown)


stof:
Note that running a blackfire profile of `vendor/bin/phpunit tests/SassSpecTest.php` on master was showing `passthru` as taking 11% of the time...

Cerdic:
This is indeed cleaner! Thanks

PR #177, opened 2020-10-13
#176 Resolve Windows-style symlinks and junctions

### LINKED ISSUE #176: Cannot resolve Windows-style directory symlinks or Windows junctions for partial path
Found in the sourcecode included in [`svewap/ws_scss`](https://github.com/svewap/ws_scss/tree/master/Resources/Private/scssphp) which resolves to v1.1.1, but after checking the current release sourcecode we think the problem has persisted.

We are using `ws_scss` to compile SCSS that consists of a base located in our site package and partials from other extensions. Running on a Windows machine, the underlying `scssphp` was not able to properly resolve a path containing a Windows-style directory symlink nor a junction.

```
`../../../../../public/typo3conf/ext/other_sitepackage/Resources/Public/Stylesheets/variables` file not found for @import: C:/path/to/repository/public/typo3conf/ext/our_sitepackage/Resources/Public/Stylesheets/style.scss on line 1, at column 0

Call Stack:
#0 import C:/path/to/repository/public/typo3conf/ext/our_sitepackage/Resources/Public/Stylesheets/style.scss (unknown file) on line 1
```

We traced back the problem to `Compiler::findImport`, where a full path is a string concatenation of the directory, a separator, and the full filename:
```
$full = $dir . $separator . $full;
```

Testing problem resolutions we could fix it by generating a `realpath` result for the concatenated string
```
$full = realpath($dir . $separator . $full);
```

Will post a pull request in a few.

I'm also linking the [related issue at `svewap/ws_scss`](https://github.com/svewap/ws_scss/issues/37).

scarcloud:
Resolves the full path of a file with `realpath`
(Resolves #176)

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -5090,13 +5090,13 @@ class Compiler
                         substr($dir, -1) !== '/' &&
                         substr($full, 0, 1) !== '/'
                     ) ? '/' : '';
-                    $full = $dir . $separator . $full;
+                    $full = realpath($dir . $separator . $full);
 
                     if (is_file($file = $full)) {
                         $found[] = $file;
                     }
                     if (! $isPartial) {
-                        $full = dirname($full) . '/_' . basename($full);
+                        $full = realpath(dirname($full) . '/_' . basename($full));
                         if (is_file($file = $full)) {
                             $found[] = $file;
                         }


Cerdic:
this patch is breaking a lot of tests as `realpath()` return false when the filename does not exists and thus no completion with partial prefix `_` or `.scss` extension is possible after that.

At least this patch would not break anything
```
diff --git a/src/Compiler.php b/src/Compiler.php
index 9eb8dec1..89fbc1df 100644
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -5092,12 +5092,12 @@ class Compiler
                     ) ? '/' : '';
                     $full = $dir . $separator . $full;

-                    if (is_file($file = $full)) {
+                    if (is_file($file = $full) || (($file = realpath($full)) && is_file($file)) ) {
                         $found[] = $file;
                     }
                     if (! $isPartial) {
                         $full = dirname($full) . '/_' . basename($full);
-                        if (is_file($file = $full)) {
+                        if (is_file($file = $full) || (($file = realpath($full)) && is_file($file)) ) {
                             $found[] = $file;
                         }
                     }
```

But I don't know if it is fixing your issue

PR #184, opened 2020-10-26
Fix the initialization of the parser for secondary entrypoints

stof:
The parse() method extract line numbers when setting the buffer, but other parse methods don't do it, while they might still trigger errors relying on that data.

This fixes tests broken by 8a30458de984e1fa79181c72378b5743907867e7 when it added parser errors in those other entrypoints.

Note that line numbers will be off for such errors. What dart-sass does is that it catches exceptions and re-throw them with an updated location when it uses the selector parser during the scss compilation. But we cannot do that easily in the current codebase. So at least, this ensures that the proper error message gets reported (even with a broken line number) instead of reporting a bug in the parser.

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -248,6 +248,7 @@ class Parser
         $this->buffer          = (string) $buffer;
 
         $this->saveEncoding();
+        $this->extractLineNumbers($this->buffer);
 
         $list = $this->valueList($out);
 
@@ -276,6 +277,7 @@ class Parser
         $this->buffer          = (string) $buffer;
 
         $this->saveEncoding();
+        $this->extractLineNumbers($this->buffer);
 
         $selector = $this->selectors($out);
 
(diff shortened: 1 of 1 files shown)


stof:
btw @Cerdic, I think you should submit your own changes through pull requests too, which would allow the CI to run *before* things get into master rather than after.

stof:
https://travis-ci.org/github/scssphp/scssphp/jobs/738236507 introduced some failures which happen only in PHP 5.6 and PHP 7.2 on Travis (the one remaining as failed tests in that PR). Could they be related to some weird edge case in the output normalization ?

stof:
@Cerdic any chance to merge this one ? It does not fix the issues on PHP 5.6 and 7.2 on CI, but it fixes the exceptions happening in all jobs (at least all jobs on recent PHP versions as older versions don't trigger a notice).

Cerdic:
This is done, thanks!

PR #188, opened 2020-10-27
Refactor Number to be immutable and spec-compliant

stof:
As part of this effort, I improved the handling of weird cases in color functions (now reporting a spec-compliant error when using units in there which are not `%`) as I had to rework usages of `$number->units()` outside the Number class, to allow making it private.

Then comes the big refactoring, where I rework all the handling of units for numbers (and so all the math) to be spec-compliant. The new code in `Number` is more or less a port of the dart-sass code of https://github.com/sass/dart-sass/blob/master/lib/src/value/number.dart (I don't implement all the "fuzzy" comparisons yet, but that might come if necessary)

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -104,7 +104,9 @@ class Compiler
 
     public static $true         = [Type::T_KEYWORD, 'true'];
     public static $false        = [Type::T_KEYWORD, 'false'];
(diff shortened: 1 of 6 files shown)


Cerdic:
I rebased manually to solve the conflict. This is integrated, thanks!

PR #190, opened 2020-10-27
PHP 8 support

adlenton:
Hi,

I took the liberty to open this small pull request which should add support for php 8 - thanks for your efforts! :)

- fixed assuming property as array when being null
- upgraded phpunit requirement for php 8

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
     "require-dev": {
         "sass/sass-spec": "2020.08.10",
         "squizlabs/php_codesniffer": "~3.5",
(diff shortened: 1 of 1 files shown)


src/Parser.php
```diff
@@ -3944,7 +3944,7 @@ private function extractLineNumbers($buffer)
     private function getSourcePosition($pos)
     {
         $low = 0;
-        $high = \count($this->sourcePositions);
+        $high = $this->sourcePositions ? \count($this->sourcePositions) : 0;
```
stof, line 3947: The right fix for that was done in #184 actually. Please revert this

Issue #189, opened 2020-10-27
Tests fail on Travis (PHP 5.6 and 7.2) on charset and null char


Cerdic:
Related to https://github.com/scssphp/scssphp/pull/184#issuecomment-716968118

Sorry for the inconvenience @stof I could indeed have used a PR here as I fully worked on a branch. 
I'm running the test with PHP 7.2 on my dev env and this is clearly working. 

It looks definitely like a weird edge case - maybe two in facts as errors in Travis PHP 5.6 seems related to charset, whereas errors on PHP 7.2 seems related to null char.

In addition the strange thing is that I had no notification from Travis to warn me of this failure, otherwise I would have had a look asap.

stof:
for the charset failure on PHP 5.6, there is a weird thing going on. the output actually looks like that:

```
content: "\\u{FFFD}";
```

Notice the double `\`. So this might be an escaping issue rather than a charset issue.

stof:
And the failing tests on PHP 7.4 builds are also related to escaping

stof:
The issue on PHP 5.6 is caused by https://3v4l.org/f2q5N
The change in https://github.com/scssphp/scssphp/commit/f7910e624a9325c227a749a8cdb921ca25e88393 only made this bug happen in a case covered by sass-spec

Cerdic:
hopefully solved by @stof with #196 and #197

PR #195, opened 2020-10-27
[Do not merge] Debug travis

stof:
Trying to understand why PHP 7.2 fail on Travis. refs #189


stof:
I managed to make it fail locally. PHP 7.2 without mbstring does not fail, but it fails with mbstring.

And this gave me the direction for the answer: the native `mb_chr` is broken on PHP 7.2 when using `mb_chr(0)`: https://3v4l.org/RgMpG

PR #196, opened 2020-10-27
Fix compat with PHP 5.6

stof:
Refs #189


stof:
the PHP 5.6 job is green :tada:

Cerdic:
Thanks fot this fix!

PR #197, opened 2020-10-27
Fix compat with PHP 7.2 with mbstring loaded

stof:
`mb_chr(0)` is broken, returning an empty string instead of a string containing the NUL byte.

Refs #189

### DIFF
--- a/src/Util.php
+++ b/src/Util.php
@@ -82,8 +82,8 @@ class Util
      */
     public static function mbChr($code)
     {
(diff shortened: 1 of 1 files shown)


stof:
@Cerdic this PR fixes the CI on PHP 7.2

So no, with #196 and this one, we can go back to green CI :tada:

Cerdic:
Thank  you so much, and sorry again for my mess!

PR #205, opened 2020-10-29
Clean the Number object

stof:
- Remove the Node parent class, as Number is more meant to represent a value than a Node, and we *never* use the properties inside that Node class for the Number object. A future PR will rework that to become `SassNumber extends Value` when working on value objects.
- Remove access to the units through the ArrayAccess BC layer as it is not used anymore. This allows to clean the constructor by removing support for the format meant only to transfer the values, as #188 actually updated all those places (I forgot to resolve the TODO at the end)


stof:
>     * A future PR will rework that to become `SassNumber extends Value` when working on value objects.

Actually, I pushed the renaming to `Value\SassNumber` in this PR too, because my work to setup phpstan as static analyser (for modern code) made me see that phpstan does not like a class named `Number` due to its `number` pseudo-type meaning `int|float`

Cerdic:
@stof @robocoder shall we make a release (as the master has a lot of fixes) before starting the refactoring?

robocoder:
Sure, let's do that.

stof:
Indeed, that makes sense

robocoder:
Done.

p.s. @stof : @Cerdic and I agreed that you should have full access to the repo; you should see a github invite.

stof:
@robocoder why adding us only on the repository rather than as organization members ?

robocoder:
That's because I could only find the membership under the repo settings. I'll look again under the organization when I get back.

stof:
@robocoder it is at https://github.com/orgs/scssphp/people

robocoder:
Huzzah! Updated the invite. Thanks.

Cerdic (commented):
@stof it looks like the spaceship operator is back in this commit
https://github.com/scssphp/scssphp/pull/205/commits/f5b4e9c1def7974076089feb137018d335600727#diff-bdd0d64c8e43471e6d175998fe2ab7d5ebe6487844bbc6315348250361b71a4bR3917

I guess this is a merge mistake? besides that seems good to me

stof:
> I guess this is a merge mistake? besides that seems good to me

indeed, that looks like a bad resolution of conflicts during a rebase

stof:
Merging this PR is on hold until the discussion on #210 is done (as we might want to do other stuff before that)

Cerdic:
I guess after this refactoring, the assertion methods
* `assertUnit()` https://github.com/scssphp/scssphp/blob/b043c42995205aeaa5a75f1937497ec5eaa9044c/src/Compiler.php#L6619
* `assertPercentOrUnitless()` https://github.com/scssphp/scssphp/blob/b043c42995205aeaa5a75f1937497ec5eaa9044c/src/Compiler.php#L6647
* `assertRange()` https://github.com/scssphp/scssphp/blob/b043c42995205aeaa5a75f1937497ec5eaa9044c/src/Compiler.php#L6676

that I introduced for #264 should move in the `Value\SassNumber` class and be public method, rewriting all the tests in the Compiler in the following form:
`$this->assertNumber($x)->asserUnit(...)`

stof:
@Cerdic I have a `value_objects` branch locally, that builds on top of that to implement the value objects for all types (I'm still missing the implementation of `SassMap` which is the most complex one due to needing to build a map with objects as keys, but based on sass equality rules, not on `spl_object_id`, making `SplObjectStorage` unsuited).
And these objects indeed have assertion methods (the same ones than in dart-sass to make maintenance easy when adding new functions). the usage would look like that: `$x->assertNumber($varName)->assertUnit($unit, $varName)` (where the only initial assumption is that `$x` is a `Value`).

Note that I'm still thinking about the migration path to avoid having to rewrite all functions in a single huge big-bang commit. I have an idea about that but I haven't tried it yet. I'll open a dedicated issue to write down my plan.

stof:
also, `assertPercentOrUnitless` *already* exists in `Number` IIRC.

stof:
I'm closing this PR. I started working on introducing value objects in a backward compatible way, by copying `Number` to `SassNumber` instead of renaming it (I plan to try rewriting `Node\Number` as a wrapper around a `Value\SassNumber` to avoid duplication but that is not done yet and might be better left for later in the migration process for performance reasons, as the compiler is not yet migrated to use value objects)

PR #207, opened 2020-10-29
Mark the units-level-3 feature as implemented

stof:
The unit table in `Number` includes the units from CSS Units Level 3


Cerdic:
Indeed!

PR #208, opened 2020-10-29
Fix the implementation of modulo

stof:
The Sass modulo works for floats and the returned value must have the same sign than the right operand (i.e. the Ruby behavior for modulo).

In PHP, `%` is the integer modulo. `fmod` must be used for float modulo.
And the PHP behavior is to is to use the sign of the left operand rather than of the right one.

### DIFF
--- a/src/Node/Number.php
+++ b/src/Node/Number.php
@@ -370,7 +370,17 @@ class Number extends Node implements \ArrayAccess
                 return NAN;
             }
 
-            return $num1 % $num2;
+            $result = fmod($num1, $num2);
+
+            if ($result == 0) {
+                return 0;
+            }
+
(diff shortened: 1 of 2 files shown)


Cerdic:
Seems right. As it will conflict with #205 I let you pick the one you want to merge first and fix the conflict then

stof:
@Cerdic I suggest merging #205 first, as I have another branch locally which is currently based on top of it to avoid conflicting, except if we want to make another release with the small fixes (that one and #207) before merging the work on refactoring values (which I started)

stof:
Thinking about that again, I will actually merge this PR first (the other one need to be fixed anyway) as I want to trigger a discussion about the path to refactor values.

PR #213, opened 2020-10-30
Fix the implementation of == and != between number and colors

stof:
Converting the number to a color and calling `opColorColor` is not OK for `==` and `!=` operators. Even when supporting color arithmetic, a number was never equal to a color.


stof:
I added tests in our own testsuite because this is not covered by sass-spec, or at least not by specs that cover only that (it might be covered by non-conformant specs where we fail multiple things)

PR #217, opened 2020-10-31
Upgrade sass-spec

stof:


tests/specs/sass-spec-exclude.txt
```diff
@@ -947,8 +1021,6 @@ libsass/debug-directive-nested/mixin
 libsass/delayed
 libsass/div
 libsass/env
-libsass/features/extend-selector-pseudoclass
-libsass/features/global-variable-shadowing
```
stof, line 951: These removals are not about increased compliance on our side. It is just that I identified the `libsass/features/*` specs as being duplicates of `core_functions/meta/feature-exists/*` ones and I removed them upstream.

tests/specs/sass-spec-exclude.txt
```diff
@@ -637,6 +701,8 @@ core_functions/selector/unify/simple/universal/and_universal/empty/and_explicit
 core_functions/selector/unify/simple/universal/and_universal/explicit/and_any
 core_functions/selector/unify/simple/universal/and_universal/explicit/and_default
 core_functions/selector/unify/simple/universal/and_universal/explicit/and_empty
+core_functions/string/index/combining_character
+core_functions/string/index/double_width_character
```
stof, line 705: These 2 failed specs are actually specs that I contributed myself upstream to cover a potential issue in our code (and this proves that we are indeed affected).

PR #229, opened 2020-11-01
Migrate CI to github actions

stof:
TODOs:

- [ ] Decide whether we keep the Travis setup
- [ ] update the branch protection config when merging.


stof:
The goal is to be able to add a job running tests with iconv but without mbstring, to help regarding https://github.com/scssphp/scssphp/issues/230. Travis does not allow us to disable mbstring as they compile it as a bundled extension. The setup-php action allows us to do it.

stof:
@robocoder @Cerdic for now, in this PR, I added all PHP versions to github actions (and not yet the job without mbstring as I prefer doing that in a separate PR as it will include fixes), which means I could migrate fully away from Travis rather than duplicating CI.
An alternative could be to use Github actions only for the job without mbstring and keeping Travis for the rest.

I would be in favor of using github actions for everything, as it keeps things in one place. Thus, it gives us the reporting about each PHP version directly in the PR statuses and we see the logs directly in checks (instead of having a single status in the PR statuses, with the status per PHP version in the checks and the logs on a separate site). What do you think ?

stof:
another nice things: github actions finishes much faster: 
image

By the time the Travis VM boots, Github actions is already done.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+                uses: actions/checkout@v2
+
+            -   name: Setup PHP
+                uses: shivammathur/setup-php@v2
+                with:
+                    coverage: "none"
+                    php-version: "${{ matrix.php }}"
+
+            -   name: Install dependencies
+                run: composer update --ansi --no-progress
```
stof, line 29: The Github Actions environment is detected as an environment not using colors by symfony/console, hence the `--ansi` option to force colors (`--colors=always` does the same for PHPUnit below as its detection logic for `--colors=auto` is basically the same). `--no-progress` avoids rendering the download progress bar, because the rendering of logs of github actions does not support the overriding of lines, and so shows all successive displays of the progress bar, which is confusing.

Compared to our Travis config, I'm using `composer update` because Composer 2 complains when using `install` without a lock file (which is turned into an `update` internally).

stof:
According to https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing#building-on-a-public-repositories-only, it seems like Travis will not be free for open-source anymore, except for repos to which they grant this on a case by case basis (a least if I understand it correctly). So that would be a reason to switch fully.

PR #232, opened 2020-11-01
Fix sourcemaps

### LINKED ISSUE #231: Sourcemaps are invalid when using UTF-8
The `@charset` directive added in https://github.com/scssphp/scssphp/commit/a82475a0b582a3ce6526ba32af56cff2e973746c as a prefix of the output is not taken into account by the sourcemap

stof:
Closes #231

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -309,8 +309,17 @@ class Compiler
 
             $out = $this->formatter->format($this->scope, $sourceMapGenerator);
 
+            $prefix = '';
+
+            if (!$this->charsetSeen && function_exists('mb_strlen')) {
+                if (strlen($out) !== mb_strlen($out)) {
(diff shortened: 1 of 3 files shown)


stof:
I tested this using the bootstrap framework, generating file sourcemaps (I hacked `bin/pscss` to support that).

When analysing the source-map with https://github.com/danvk/source-map-explorer, it reports 22% of unmapped content without the fix in the formatter to map all lines (disabling border checks due to the issue of adding a mapping when writing `\n`, but the percentage stays the same when running with only this issue patched and applying border checks in the analysis).
After the patch, it reports only 17 bytes as unmapped (which is exactly the `@charset "UTF-8";` prefix on the first line)

stof:
oh, and the fix for multiline writes was found by looking at how dart-sass builds its source maps.

PR #238, opened 2020-11-02
Update build badge per #229

robocoder:


stof (approved):
Indeed, I totally forgot that

PR #239, opened 2020-11-02
Expose SassScriptException as a non internal class

stof:
When I introduced `SassScriptException` in #188, I tagged it as `@internal`. But that was a mistake. For developers *using* the library, this exception is indeed an internal implementation detail that they will never see. But for developers *extending* SassScript with custom functions, this exception is absolutely not an internal implementation detail. It should actually be the go-to exception for custom function rather than relying on `Compiler::error()`. It avoids needing access to the Compiler to trigger errors (which is why `Number` uses it, and why `SassScriptException` exists in `dart-sass` too, where I got the idea).

### DIFF
--- a/src/Exception/SassScriptException.php
+++ b/src/Exception/SassScriptException.php
@@ -3,11 +3,12 @@
 namespace ScssPhp\ScssPhp\Exception;
 
 /**
- * Internal exception thrown in places not having access to reporting the location.
+ * An exception thrown by SassScript.
  *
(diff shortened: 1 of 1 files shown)


stof:
Looking forward at ScssPhp 2.x, the plan is to have all (or at least most) of SassScript implemented outside the Compiler class, by having things in methods on the Value objects and Sass functions defined in separate classes. So `SassScriptException` would be used a lot more then.

Issue #242, opened 2020-11-02
Imports are not spec compliant


stof:
The [official spec about imports](https://sass-lang.com/documentation/at-rules/import#finding-the-file) describes this algorithm:

1. try to resolve the file relatively to the current file
2. if failed, try to resolve the file using load paths (what we call "import paths" in scssphp)

scssphp does not implement that at all. It never resolves files relatively. What it does instead is prepending some load paths in the configured list when entering an imported file and removing it when leaving. And by default, it puts the empty string as first load path in the list (which leads to resolving to `getcwd()` due to the way PHP resolves relative paths). But this one can be removed (and currently, `pscss -I /my/load/path` removes it)
This has several drawbacks:
- relative imports for the root stylesheet work only if you take care of doing `chdir(dirname($path));` before the compilation or if you add it as a load path. this is quite inconvenient
- once you start importing files, you leak load paths for the imports done by these files. this allows a file to resolve imports relatively to its importing file (and the importing file of the importing file, etc...)
- if you catch a compilation exception and then reuse the Compiler for another compilation, these leaked load paths are still there and are the first ones...
- by doing `$compiler->setImportPaths([]);`, you can effectively disable `@import` as there won't be any resolution done (and this is even [documented](https://github.com/scssphp/scssphp/blob/51e7484d2b590031ffa5fbabddea7682e534c88b/docs/docs/index.md#import-paths)

see https://github.com/sass/sass-spec/pull/1589 for a spec failing on this.

robocoder:
I think this stems from originally having the Server class as part of the library, and to some degree supporting stdin when using `bin/pscss`

stof:
well, for `stdin`, using the current working directory as location of the root file is fine (just like passing a relative file as input gets resolved based on `cwd`). But that should not make the CWD a load path for the whole compilation.

PR #243, opened 2020-11-02
Fix the registration of load paths in pscss

stof:
By making calls to `addImportPath` rather than using `setImportPaths`, the import path based on the current working directory is preserved. While this does not ensure spec compliance for imports (see #242), it makes relative imports mostly work.

I discovered that while trying to use the official sass-spec running with `-c /path/to/bin/pscss`, as that was easier to run a spec that I was working on in my sass-spec clone (the runnner passes the root path of specs as a load path). With this patch applied, such workflow works.

### DIFF
--- a/bin/pscss
+++ b/bin/pscss
@@ -178,7 +178,9 @@ if ($dumpTree) {
 $scss = new Compiler();
 
 if ($loadPaths) {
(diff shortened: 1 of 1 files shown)


stof:
This is not necessary anymore if we merge #244

PR #244, opened 2020-11-03
Refactor the resolution of imports to be spec compliant 

### LINKED ISSUE #242: Imports are not spec compliant

stof:
Closes #242

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -133,7 +133,7 @@ class Compiler
     /**
      * @var array<int, string|callable>
      */
-    protected $importPaths = [''];
+    protected $importPaths = [];
     /**
      * @var array<string, Block>
      */
@@ -266,6 +266,20 @@ class Compiler
     protected $callStack = [];
 
     /**
+     * The directory of the currently processed file
+     *
+     * @var string
+     */
+    private $currentDirectory;
+
+    /**
+     * The directory of the input file
+     *
+     * @var string
+     */
+    private $rootDirectory;
+
+    /**
      * Constructor
      *
      * @param array|null $cacheOptions
@@ -359,6 +373,13 @@ class Compiler
         $this->charsetSeen    = null;
         $this->shouldEvaluate = null;
         $this->ignoreCallStackMessage = false;
+
+        if (!\is_null($path) && is_file($path)) {
+            $this->currentDirectory = dirname(realpath($path) ?: $path);
+        } else {
+            $this->currentDirectory = getcwd();
+        }
+        $this->rootDirectory = $this->currentDirectory;
 
         try {
             $this->parser = $this->parserFactory($path);
@@ -3093,7 +3114,7 @@ class Compiler
             case Type::T_DEBUG:
                 list(, $value) = $child;
 
-                $fname = $this->sourceNames[$this->sourceIndex];
+                $fname = $this->getPrettyPath($this->sourceNames[$this->sourceIndex]);
                 $line  = $this->sourceLine;
                 $value = $this->compileDebugValue($value);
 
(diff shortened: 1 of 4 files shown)


src/Compiler.php
```diff
@@ -5336,13 +5336,124 @@ public function findImport($url)
...
         }
 
-        if ($urls) {
-            if (! $hasExtension || preg_match('/[.]scss$/', $url)) {
-                throw $this->error("`$url` file not found for @import");
-            }
+        throw $this->error("`$url` file not found for @import");
+    }
+
+    private function resolveImportPath($url, $baseDir)
```
stof, line 5342: This method (and the following helper methods) are implementing the logic at https://github.com/sass/dart-sass/blob/28a582f45e3f9f3384071a83db2012b4e2e16707/lib/src/importer/utils.dart#L41-L44 with 2 exceptions (not counting the fact that dart-sass makes the `path.join()` in the caller):

- support for the `.sass` extension, as we don't support parsing them (this means we don't consider them as a source of conflict either)
- support for `.import.scss` files, as that's part of implementing Sass modules (this one will become support when/if we implement modules, but the code will probably have changed quite a bit at that point anyway)

stof:
Regression tests for our broken implementation of relative imports have been accepted in sass-spec itself: https://github.com/sass/sass-spec/pull/1589 :tada:

src/Compiler.php
```diff
@@ -5336,13 +5336,124 @@ public function findImport($url)
...
+    }
+
+    /**
+     * @param string $path
+     *
+     * @return string[]
+     */
+    private function tryImportPath($path)
+    {
+        $partial = dirname($path).'/_'.basename($path);
```
Cerdic, line 5408: Is it the expected behaviour that `@import '_partial';`  will try to import  `__partial.scss`?

I had the feeling that an explicit partial reference in the `@import` should just be taken as is, not trying to partialize the already partial file (at least that's what scssphp was doing before).
Am I wrong?

stof: dart-sass will do it too.

Cerdic: yep seems more a flaw in the design than a real feature :p 
but if dart-saas do that, let's stick on it

This seems all good for me besides that.

stof: I opened https://github.com/sass/sass/issues/2940 to clarify that.

stof: @Cerdic it's `dart-sass`, not `dart-saas` :smile:

stof: The Sass team answered that this behavior is the same in dart-sass and ruby-sass, and not worth a BC break to undo it, even though they would have preferred doing that differently retrospectively (that was probably not intended). So I think it is good we follow the same rule.

Issue #246, opened 2020-11-03
test suite is broken on master


Cerdic:
on b4d5ecfb

```
1) ScssPhp\ScssPhp\Tests\SassSpecTest::testTests with data set "16/5452. core_functions/color/adjust_color/error/hsl_and_rgb/blue_and_lightness" ('16/5452. core_functions/color...htness', array('', 'a {b: adjust-color(red, $blue...)}\n\n', array(), 'blue_and_lightness'), array('', '', 'Error: RGB parameters may not...et\n\n', array()))
Failed asserting that exception of type "ScssPhp\ScssPhp\Exception\SassException" is thrown.

2) ScssPhp\ScssPhp\Tests\SassSpecTest::testTests with data set "17/5452. core_functions/color/adjust_color/error/hsl_and_rgb/green_and_saturation" ('17/5452. core_functions/color...ration', array('', 'a {b: adjust-color(red, $gree...)}\n\n', array(), 'green_and_saturation'), array('', '', 'Error: RGB parameters may not...et\n\n', array()))
Failed asserting that exception of type "ScssPhp\ScssPhp\Exception\SassException" is thrown.

3) ScssPhp\ScssPhp\Tests\SassSpecTest::testTests with data set "18/5452. core_functions/color/adjust_color/error/hsl_and_rgb/red_and_hue" ('18/5452. core_functions/color...nd_hue', array('', 'a {b: adjust-color(red, $red:...)}\n\n', array(), 'red_and_hue'), array('', '', 'Error: RGB parameters may not...et\n\n', array()))
Failed asserting that exception of type "ScssPhp\ScssPhp\Exception\SassException" is thrown.

4) ScssPhp\ScssPhp\Tests\SassSpecTest::testTests with data set "129/5452. core_functions/color/change_color/error/hsl_and_rgb/blue_and_lightness" ('129/5452. core_functions/colo...htness', array('', 'a {b: change-color(red, $blue...)}\n\n', array(), 'blue_and_lightness'), array('', '', 'Error: RGB parameters may not...et\n\n', array()))
Failed asserting that exception of type "ScssPhp\ScssPhp\Exception\SassException" is thrown.

[...]

PR #248, opened 2020-11-03
Improve the Makefile to ensure vendors are uptodate

stof:
Refs #246 

Running `make test` or `make sass-spec` will not automatically run `composer update` if the composer.json was modified since the last time the vendor folder was updated. This avoids forgetting it when a PR is merged to upgrade sass-spec for instance.

For contributors not using `make`, this won't help of course. They will still need to remember running `composer update`, just like the current status quo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,12 @@
-install-test:
-	mkdir -p vendor/sass
-	cd vendor/sass && git clone https://github.com/sass/sass-spec.git && cd ../..
-
-test:
+test: vendor
 	vendor/bin/phpunit --colors tests
 
-sass-spec:
+sass-spec: vendor
 	TEST_SASS_SPEC=1 vendor/bin/phpunit --colors tests 2>&1 | tee /tmp/sass-spec.log | tail -2
 
-standard:
+standard: vendor
 	vendor/bin/phpcs -s --standard=PSR12 --exclude=PSR12.Properties.ConstantVisibility --extensions=php bin src tests *.php
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,12 +1,12 @@
-install-test:
```
stof, line 1: I'm dropping that one as relying on composer to install sass-spec is better: it ensures the right commit of sass-spec is used instead of taking the last one at the time you ran `make install-test`.
Thus, `make install-test` was probably broken if `sass-spec` is already installed.

Makefile
```diff
@@ -1,12 +1,12 @@
...
+sass-spec: vendor
 	TEST_SASS_SPEC=1 vendor/bin/phpunit --colors tests 2>&1 | tee /tmp/sass-spec.log | tail -2
 
-standard:
+standard: vendor
 	vendor/bin/phpcs -s --standard=PSR12 --exclude=PSR12.Properties.ConstantVisibility --extensions=php bin src tests *.php
+
+vendor: composer.json
+	composer update
+	touch $@
```
stof, line 12: this `touch vendor` ensures that the modified date of the `vendor` folder itself is updated. This way, `make` will detect that `vendor` is uptodate when you run `make test` and won't run the command again (nice as you will run tests much more often than changing the `composer.json` when working on features)

stof:
Currently, this relies on `composer` being available in the PATH. I think it is a fine assumption, but `make` has features to make that configurable if needed.

PR #252, opened 2020-11-04
[wip] Support for HWB colors en functions

Cerdic:
Add the support for HWB color representations and thus
- add `hwb()`, `whiteness()` and `whiteness()` functions
- add support for `$whiteness` and `$blackness` argument in `adjust-color()`, `change-color()` and `scale-color()`

I had to tweak a bit the Sass-spec execution to translate `color.hwb()` `color.whiteness()` and `color.whiteness()` calls from the color module and supress the call of the color module when it was not uses for something else.

Thus around 100 more specs are played and compliant around this functions.

Question is that we are adding support of direct call to `hwb()`, `whiteness()` and `whiteness()` without prefixing with the `color.` module and then this is not fully compliant.

Wonder if 
- we can integrate as is
- we should wait to be able to support module support
- introduce some syntax derogation to support `color.` prefixed color functions without supporting modules

Keeping this in standby meanwhile, but I could improve this PR by adding a lot of argument check in the function colors as I did for the hwb here.


stof:
> Question is that we are adding support of direct call to `hwb()`, `whiteness()` and `whiteness()` without prefixing with the `color.` module and then this is not fully compliant.

Sass has explicitly decided *not* adding these as global functions to avoid issues. The CSS spec for `hwb()` is not fully stable yet (and not supported by browsers), so creating a a global Sass function shadowing the CSS function could lead to CSS incompatibilities. See https://github.com/sass/sass/blob/master/accepted/color-4-hwb.md#design-decisions


> Wonder if
> 
>     * we can integrate as is
>     * we should wait to be able to support module support
>     * introduce some syntax derogation to support `color.` prefixed color functions without supporting modules

My vote for now would be to wait until we have module support, to avoid spec incompatibilities (can be reconsidered later of course).

Syntax derogation for color module functions would be a fragile hack. There is nothing forbidding you to import `sass:color` under a different alias than `color`. And it would force us to introduce a syntax derogation for `@use "sass:color"` too, as writing the file without it would be invalid sass that dart-sass cannot parse.

> Keeping this in standby meanwhile, but I could improve this PR by adding a lot of argument check in the function colors as I did for the hwb here.

Given that our handling of SassScript types is quite loose regarding compliance in the parser/compiler, I fear that introducing stricter type checks in functions in 1.x could be risky (I'm not sure colors are always represented as `T_COLOR` in 100% of cases. I think there is some magic allowing them to be represented in different ways that get converted at some point, and then we would have to make sure that the type check always happens after the magic and not before. And our magic is not spec compliant anyway as it allows coercing values between types while Sass does not support that).

I'd rather delay it until 2.0 where my refactoring of Value will enforce type checks in all functions.

stof:
> Thus around 100 more specs are played and compliant around this functions.

This is actually a lie. We are not compliant with the specs, as we edit them before running them (and so we are still unable to run the official specs)

Cerdic:
That was also my feeling when finishing the work. 

Let's keep it as a draft for now and we'll see  later for this integration, maybe partial: what we could probably do is to integrate only the underlying conversion RGB<->HWB and the support for the `$whiteness`/`$blackness` args in the color function transformations
(which was my starting point on this work, seeing a lot of new errors on the sass-spec test due to the non support of this arguments)

Cerdic:
I'm closing this one as #264 is an alternative proposal

PR #257, opened 2020-11-07
Fix usage with iconv but not mbstring

### LINKED ISSUE #230: The tests are failing with iconv but without mbstring
Related to https://github.com/scssphp/scssphp/issues/203

Supporting iconv without mbstring should be simple to fix and so we would not deprecate such setup.
But a proper fix should involve having CI for that.

stof:
Closes #230

The CI will now ensure this keeps working.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,13 +7,18 @@ on:
 
 jobs:
     tests:
-        name: Tests
+        name: "Tests on PHP ${{ matrix.php }}${{ matrix.name_suffix }}"
         runs-on: ubuntu-latest
 
         strategy:
             fail-fast: false
             matrix:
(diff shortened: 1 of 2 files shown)


stof:
As part of that PR, I reworked the name of jobs,  to control myself how the different jobs of the matrix are named.

.github/workflows/ci.yml
```diff
@@ -7,13 +7,18 @@ on:
...
     tests:
-        name: Tests
+        name: "Tests on PHP ${{ matrix.php }}${{ matrix.name_suffix }}"
         runs-on: ubuntu-latest
 
         strategy:
             fail-fast: false
             matrix:
                 php: [ '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0' ]
+                extensions: ''
```
stof, line 17: Specifying `extensions` here ensures that the extra job added with a custom value for `extensions` does not swallow the normal job by matching its values.

Issue #268, opened 2020-11-10
Improve the compilation cache invalidation options


Cerdic:
Linked to https://github.com/scssphp/scssphp/issues/256#issuecomment-724747705

There is now 2 usages of the Cache in ScssPhp : 
- in the Parser https://github.com/scssphp/scssphp/blob/master/src/Parser.php#L236 to get the result of an already parsed content
- in the Compiler https://github.com/scssphp/scssphp/blob/master/src/Compiler.php#L346 to get the result of an already Compiled content

For the Parser, the invalidation is straightforward: the result only rely on the content itself, and if content is unchanged it's safe to rely on the parsed result in cache.

For the Compiler, the invalidation is a bit more tricky. Even if the content of file is unchanged, the compilation result can change if one `@import`ed file is modified, which can happen in 2 cases:
1. the file that was imported in the previous compilation has changed, so it need to be re-parsed and then the whole parent file need to be compiled again
2. the `@import` resolving itself doesn't produce the same result (because the path changed, or because a new file is now available on the disk in a higher priority part of the path - ie overriding)

The first case is adressed by checking the timestamp of files that where used to produce the compiled cache: in the cache we are storing all imported files, and thus the Compiler can check if any of the files has been modified since the compilation result was stored in cache https://github.com/scssphp/scssphp/blob/master/src/Compiler.php#L350

The second case is not adressed at all at the moment.
A different path betwen the cached compilation result and the present call could be easy to manage (and I have to confess I didn't included that in ScssPHP because I was managing this issue in my calling function).

Checking that all `@import` would resolve the same way to detect the availability of a new file needs more informations to be stored in the cache:
- basically we need to register every call to `findImport()` with the `$url` argument and the `$this->currentDirectory` context and the result, store this list in the cache
- then when checking for the validity of a cached Compile result, go through this list, and check that all the `findImport()` would give the same final result
- if one of the `findImport()` gives a different filepath as a result OR if the timestamp changed on this file, then the cached Compile result should be ignored.

This process can be time consuming cause needing a lot of disk access, thus it should be an option for the Compiler to make a fast check (as now, only on the timestamp) or a deep check (checking all `findImport()`)

mahagr:
Looks like we are on the same page (or you did read my mind). See my comment in the linked discussion. :)

stof:
> For the Parser, the invalidation is straightforward: the result only rely on the content itself, and if content is unchanged it's safe to rely on the parsed result in cache.

[...]

Issue #222, opened 2020-11-01
[RFC] Stateless API for the compiler


stof:
Currently, the Compiler is a stateless class due to its `getParsedFiles` getter, which allows to get some details about the previous compilation.
This stateful getter means that the Compiler has to keep that info around. It also means that some weird things might happen if the same `Compiler` instance is shared between multiple part of a project, and that one of them expects to be able to call this getter until the next time it triggers a compilation, while giving back the execution to other code (which could then potentially trigger a different compilation).

I suggest an alternate signature (using PHP types for brevity here, but that would be implemented in phpdoc instead due to the support for PHP 5.6):

```php
<?php

class Compiler
{
    public function compile(string $code, ?string $path = null): CompilationResult
    {
        // ...
    }
}

class CompilationResult
{
    public function getCss(): string
    {
        // ...
    }

    // name matching the JS API
    public function getIncludedFiles(): array
    {
        // ...
    }

    public function __toString(): string
    {
        return $this->getCss(); // To reduce the impact of the BC break
    }

    // The sourceMap content, if it was generated
    public function getMap(): ?string
    {
        // ...
    }
}
```

This result object is similar to what the JS API does (except they have a included files in a nested `stats` object, which I decided to skip as relevant).

mahagr:
I like this approach -- in fact, I just mentioned about this behavior in #214, where I mention fixing it in our class override.

Using the same Compiler instance for compiling multiple files can really speed up compile times as the shared includes are already cached. It is possible to do this already with current version of the compiler, but not without minor fixes and hacks to make it to work properly.

Cerdic:
seems good to me as well

Cerdic:
Fixed by #270

PR #270, opened 2020-11-11
Refactor compilation result

Cerdic:
This is an implementation of #222 and fixing #268 in a efficient way

As a matter of fact, there is no BC break.
The CompilationResult is storing 3 differents lists:
- parsedFiles, as it was in the Compiler Class, listing the parsed files with their timestamp
- importedFiles which list all the findImport resolved as a file, with the current dir and search path. This is used for a deep check of cache validity
- includedFiles which list all the `@import` path, whether it's resolved as a file (then it is in the importedFiles as well) or a just kept as a css `@import` directive. We are not using this list, but this is an output of dart-sass and while refactoring I tried to stay as close as possible.

I think that importedFiles and parsedFiles could probably be merged, but I didn't had a deep check on that and prefered to not introduce any breaking in this refactoring. 
This is probably an optimisation for later.

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -425,8 +425,7 @@ class Compiler
                 }
             }
 
-            $this->compilationResult->setCss($prefix);
-            $this->compilationResult->appendCss($out);
+            $this->compilationResult->setCss($prefix . $out);
 
             if (! empty($out) && $this->sourceMap && $this->sourceMap !== self::SOURCE_MAP_NONE) {
                 $sourceMap    = $sourceMapGenerator->generateJson($prefix);
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -211,6 +211,11 @@ class Compiler
     protected $cache;
 
     /**
+     * @var bool
+     */
+    protected $cacheCheckImportResolutions = false;
+
+    /**
      * @var int
      */
     protected $indentLevel;
@@ -295,6 +300,9 @@ class Compiler
 
         if ($cacheOptions) {
             $this->cache = new Cache($cacheOptions);
+            if (!empty($cacheOptions['checkImportResolutions'])) {
+                $this->cacheCheckImportResolutions = true;
+            }
         }
 
         $this->stderr = fopen('php://stderr', 'w');
@@ -350,8 +358,11 @@ class Compiler
             $compileOptions = $this->getCompileOptions();
             $cache          = $this->cache->getCache('compile', $cacheKey, $compileOptions);
 
-            if (!empty($cache) && $cache instanceof CompilationResult && $cache->checkValid()) {
-                return $cache;
+            if (!empty($cache) && $cache instanceof CompilationResult) {
+                $cache->setIsCached(true);
+                if ($cache->checkValid($this->cacheCheckImportResolutions, $this)) {
+                    return $cache;
+                }
             }
         }
 
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -224,11 +224,6 @@ class Compiler
     protected $extendsMap;
 
     /**
-     * @var string[]
-     */
-    protected $importedFiles = [];
-
-    /**
      * @var CompilationResult
      */
     protected $compilationResult;
@@ -2472,10 +2467,10 @@ class Compiler
         if ($rawPath[0] === Type::T_STRING) {
             $path = $this->compileStringContent($rawPath);
 
-            if (strpos($path, 'url(') !== 0 && $path = $this->findImport($path)) {
-                if (! $once || ! \in_array($path, $this->importedFiles)) {
-                    $this->importFile($path, $out);
-                    $this->importedFiles[] = $path;
+            if (strpos($path, 'url(') !== 0 && $filePath = $this->findImport($path, $this->currentDirectory)) {
+                if (! $once || ! \in_array($filePath, $this->compilationResult->getImportedFiles())) {
+                    $this->importFile($filePath, $out);
+                    $this->compilationResult->addImportedFile($this->currentDirectory, $path, $filePath);
                 }
 
                 return true;
@@ -5316,10 +5311,11 @@ class Compiler
      * @api
      *
      * @param string $url
+     * @param string $currentDir
      *
      * @return string|null
      */
-    public function findImport($url)
+    public function findImport($url, $currentDir = null)
     {
         // for "normal" scss imports (ignore vanilla css and external requests)
         // Callback importers are still called for BC.
@@ -5341,8 +5337,8 @@ class Compiler
             return null;
         }
 
-        if (!\is_null($this->currentDirectory)) {
-            $relativePath = $this->resolveImportPath($url, $this->currentDirectory);
+        if (!\is_null($currentDir)) {
+            $relativePath = $this->resolveImportPath($url, $currentDir);
 
             if (!\is_null($relativePath)) {
                 return $relativePath;
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -13,6 +13,7 @@
 namespace ScssPhp\ScssPhp;
 
 use ScssPhp\ScssPhp\Base\Range;
+use ScssPhp\ScssPhp\Compiler\CompilationResult;
 use ScssPhp\ScssPhp\Compiler\Environment;
 use ScssPhp\ScssPhp\Exception\CompilerException;
 use ScssPhp\ScssPhp\Exception\ParserException;
@@ -140,10 +141,7 @@ class Compiler
      * @var array<string, Block>
      */
     protected $importCache = [];
-    /**
-     * @var string[]
-     */
-    protected $importedFiles = [];
+
     protected $userFunctions = [];
     protected $registeredVars = [];
     /**
@@ -224,10 +222,17 @@ class Compiler
      * @var array<string, int[]>
      */
     protected $extendsMap;
+
     /**
-     * @var array<string, int>
+     * @var string[]
      */
-    protected $parsedFiles;
+    protected $importedFiles = [];
+
+    /**
+     * @var CompilationResult
+     */
+    protected $compilationResult;
+
     /**
      * @var Parser|null
      */
@@ -290,7 +295,7 @@ class Compiler
      */
     public function __construct($cacheOptions = null)
     {
-        $this->parsedFiles = [];
+        $this->compilationResult = new CompilationResult();
         $this->sourceNames = [];
 
         if ($cacheOptions) {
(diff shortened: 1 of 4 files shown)


Cerdic:
> As a matter of fact, there is no BC break.

this is not true because I added an argument to `findImport()` to provide it the current directory instead of using the property, and I'm using this argument in the Compiler, which means a break if someone had overiden this method.

We could avoid that by not using this argument in the Compiler itself (and then fallbacking to the property if no arg provided to the method) and using it only in the Cache checking. 
This means without any changement everything should work as before, but if you want to use the new cache validation option, you have to also change your `findImport()` override.

This is a bit less clean for the `findImport()` method itself, but seems acceptable.

src/Compiler.php
```diff
@@ -5054,11 +5067,12 @@ public function addParsedFile($path)
      *
      * @api
      *
+     * @deprecated
```
stof, line 5070: we should add the `@trigger_error('...', E_USER_DEPRECATED);` to report the deprecation.

src/Compiler/CompilationResult.php
```diff
@@ -0,0 +1,305 @@
...
+class CompilationResult
+{
+    protected static $resolvedImport = [];
+
+    protected $isCached = false;
+
+    /**
+     * @var string
+     */
+    protected $css = '';
```
stof, line 31: I suggest using `private` visibility in the new class, to avoid any BC needs on the properties.

src/Compiler/CompilationResult.php
```diff
@@ -0,0 +1,305 @@
...
+     */
+    public function setCss($css)
+    {
+        $this->css = $css;
+    }
+
+    /**
+     * @param string $css
+     */
+    public function appendCss($css)
```
stof, line 86: AFAICT, this method is only used  to manage the prefix, where you even set the prefix as CSS and then append the actual CSS with this method. I'd rather not have this method here, and keep the concatenation in the compiler class.

src/Compiler/CompilationResult.php
```diff
@@ -0,0 +1,305 @@
...
+     *
+     * @api
+     *
+     * @param string $path
+     *
+     * @return void
+     */
+    public function addParsedFile($path)
+    {
+        if (isset($path) && is_file($path)) {
```
stof, line 110: this `isset` is useless useless if the param is really a `string` and not `string|null`

src/Compiler/CompilationResult.php
```diff
@@ -0,0 +1,305 @@
...
+
+    public function __toString()
+    {
+        return $this->getCss(); // To reduce the impact of the BC break
+    }
+
+
+    /**
+     * Store the sourceMap and its storage data
+     * @param $sourceMap
```
stof, line 196: missing type

src/Compiler/CompilationResult.php
```diff
@@ -0,0 +1,305 @@
...
+        } else {
+            $this->sourceMapFile = $sourceMapFile;
+            $this->sourceMapUrl = $sourceMapUrl;
+        }
+    }
+
+    /**
+     * The sourceMap content, if it was generated
+     *
+     * @return string
```
stof, line 214: should be `string|null`

src/Compiler/CompilationResult.php
```diff
@@ -0,0 +1,305 @@
...
+use ScssPhp\ScssPhp\Util;
+use ScssPhp\ScssPhp\Exception\CompilerException;
+
+/**
+ * Compiler environment
+ *
+ */
+class CompilationResult
+{
+    protected static $resolvedImport = [];
```
stof, line 24: why static ?

Cerdic: Static is used here for performance issue accros multiples file compilation sharing mixins or functions libs.

My testing use case is 4 .scss compilation: the full BS4 as a first and then 3 others using parts of BS4 (like variables, mixins..). Without the static a lot of validation are occuring multiple times and this is costly.

Using this static allows a total cost for this full check that is quite acceptable.

I guess to be bulletproof, the `checkValid()` method should have an optional `$forceRefresh` argument to empty the static and force a full re-check?

stof: if we need to share some state between compilations, I think this should live in the Compiler class rather than being in a static property in CompilationResult. Each compilation has its own CompilationResult.

btw, using a static property here would even leak the shared state between multiple Compiler instances, which might have different configuration.

Cerdic: This is purely an internal storage stuff and the case of multiple compiler instances with different configurations is taken in account through the hash, to reuse previous check only if 
https://github.com/scssphp/scssphp/pull/270/commits/f57610d92bcfbbf50e387d64a950e7d95ff037c1#diff-a95b369f4118d2d8d85bb94f822b8d5ca7e92c93a9c0d533e43742b8f3e6e742R277

In my test case indeed each scss compilation occurs with a new Compiler Instance - but I guess I could now use a static instance to reuse it accross several compilations.

Alternatively I can remove the static optimization from the CompilerResult class, delivering something ready to use and 100% clean - even if not the fastest implementation.
And besides that add an option to the `getImportedFiles()` method to return the full array and not just one column.

Then it allows users to implement their own checking method, possibly faster with some compromise on compilation isolation.

For me this is making a difference:
- the fast checking method, with only the timestamps is around **~10ms** for the whole scss compilations
- with a deep checking with non-static array optimisation : **~85ms**
- with a deep checking with static array optimisation :  **~45ms**

But I also have to admit the difference is not as big as I had with my first implementation where i measured a total of  ~250ms with a full check (but I can't figure out what I changed since).

Let's go for this compromise?

src/Compiler/CompilationResult.php
Cerdic: In fact it's `string|null` and I prefer to keep this as is as I don't want to introduce a test here 
https://github.com/scssphp/scssphp/blob/master/src/Compiler.php#L481
which would tend to make someone move the previous line in the if whereas we need to fill the array whatever the value of `$path`

src/Compiler.php
Cerdic: fixed

src/Compiler/CompilationResult.php
Cerdic: private on all the properties, fixed

src/Compiler/CompilationResult.php
Cerdic: removed

src/Compiler/CompilationResult.php
Cerdic: fixed

src/Compiler/CompilationResult.php
Cerdic: fixed

src/Compiler/CompilationResult.php
Cerdic: so, no more static, just a local variable

src/Compiler/CompilationResult.php
stof: please change it to a null check then to be clearer

src/Compiler.php
```diff
@@ -2472,16 +2477,18 @@ protected function compileImport($rawPath, OutputBlock $out, $once = false)
...
+                    $this->compilationResult->addImportedFile($this->currentDirectory, $path, $filePath);
                 }
 
                 return true;
             }
 
-            $this->appendRootDirective('@import ' . $this->compileImportPath($rawPath) . ';', $out);
+            $path = $this->compileImportPath($rawPath);
+            $this->appendRootDirective('@import ' . $path . ';', $out);
+            $this->compilationResult->addIncludedFile($path);
```
stof, line 2491: this looks wrong to me. This file is not included as it is compiled to a CSS `@import` (and might not even be a file that exists on disk)

src/Compiler.php
```diff
@@ -2494,7 +2501,9 @@ protected function compileImport($rawPath, OutputBlock $out, $once = false)
 
             foreach ($rawPath[2] as $path) {
                 if ($path[0] !== Type::T_STRING) {
-                    $this->appendRootDirective('@import ' . $this->compileImportPath($rawPath) . ';', $out);
+                    $path = $this->compileImportPath($rawPath);
+                    $this->appendRootDirective('@import ' . $path . ';', $out);
+                    $this->compilationResult->addIncludedFile($path);
```
stof, line 2506: this is not an included file either

src/Compiler.php
```diff
@@ -2507,7 +2516,9 @@ protected function compileImport($rawPath, OutputBlock $out, $once = false)
             return true;
         }
 
-        $this->appendRootDirective('@import ' . $this->compileImportPath($rawPath) . ';', $out);
+        $path = $this->compileImportPath($rawPath);
+        $this->appendRootDirective('@import ' . $path . ';', $out);
+        $this->compilationResult->addIncludedFile($path);
```
stof, line 2521: this is not an included file either.

src/Compiler.php
```diff
@@ -5066,11 +5080,14 @@ public function addParsedFile($path)
      *
      * @api
      *
+     * @deprecated
      * @return array
      */
     public function getParsedFiles()
     {
-        return $this->parsedFiles;
+        @trigger_error('addParsedFile() is no longer a method from Compiler but from CompilationResult.'
```
stof, line 5088: wrong method name

src/Compiler/CompilationResult.php
```diff
@@ -0,0 +1,292 @@
...
+    public function getIncludedFiles()
+    {
+        return $this->includedFiles;
+    }
+
+    // A map from source file URLs to the corresponding [SourceFile]s.
+    //
+    // This can be passed to [sourceMap]'s [Mapping.spanFor] method. It's `null`
+    // if source mapping was disabled for this compilation.
+    public function getSourceFiles()
```
stof, line 172: why do we have such method ?

stof:
I'm actually wondering whether this caching layer doing complex IO-based invalidation (as it has to resolve all imports again) makes sense or whether we should rather drop it. dart-sass does not have any such feature.

src/Compiler.php
Cerdic: I will dig again, but as long I can remember I sticked to what dart-saas was doing on this files liste, even if it doesn't feels what you are expecting.

stof: dart-sass is only adding values in this file list for sass imports, not for CSS imports (which are compiled by `visitStaticImport`)

stof: the doc about filesystem import vs URI is related to the fact that dart-sass supports using URIs for imports, not only paths (when running the dart API, it can support `package:` URIs which resolve to a dart package installed in the project for instance)

Issue #179, opened 2020-10-21
[RFC] Upstream our tests to the sass-spec repo


stof:
Currently, this repo contains a bunch of tests in `tests/input`. These tests are not guaranteed to be compatible with the official sass implementation (see for instance https://github.com/scssphp/scssphp/commit/a16c1965f6cebcd43abdfba8a2e4149d657222a0 where one of our tests was explicitly requiring an output incompatible with the official spec).
But we also have tests which cover cases not represented in upstream specs (for instance, the fix in https://github.com/scssphp/scssphp/commit/237bfd734b8281b13bc13af855f0cfbefe6c6ad7 does not unlock any upstream tests). Some tests are also added as regressions tests for reported bugs (https://github.com/scssphp/scssphp/commit/07155c1a0f05b0f263a99beec61db2999b1f38a3 for instance), but might not match upstream behavior.

I think it would be great to work on upstreaming tests we have (this would require identifying what each input was meant to cover, due to the strict rules of sass-spec about having tests targetted on a single thing at a time). This way, we would guarantee that our behavior for these edge cases matches the official sass behavior.

stof:
#670 is a good first step in that direction by first making our tests granular first (as only granular tests can be upstreamed due to the upstream requirements)

PR #274, opened 2020-11-17
Fix #273 and escaped selector sass-spec test

### LINKED ISSUE #273: Breaking change un 1.3 on selector re-parsing (again :/)
Related to #245 and compiled selector re-parsing to check they are valid

```scss
/// Utility mixin for grid.
/// @param {list} $gutters Column and row gutters (default is 40px).
/// @param {string} $breakpointName Optional breakpoint name.
@mixin grid($gutters: 40px, $breakpointName: null) {

	// Cells.
		$x: '';
		@if $breakpointName {
			$x: '\\28' + $breakpointName + '\\29';
		}
		.\31 2u#{$x}, .\31 2u\24#{$x} { width: 100%; clear: none; margin-left: 0; }
		.\31 2u\24#{$x} + *,
		.\31 u\24#{$x} + * {
			clear: left;
		}
		.\-11u#{$x} { margin-left: 91.6666666667% }
}
	
.test {
    @include grid(40px, 'xlarge');
}
```

produce
```
`$\28xlarge\29 + *` is not a valid Selector in `.\312u\24\28xlarge\29 + *, .\31u$\28xlarge\29 + *`: failed at `$\28xlarge\29 + *` ScssPhp\ScssPhp\Compiler::evalSelectors on line 1, at column 32
```

Whereas Sassmeister produce

```css
.test .\31 2u\(xlarge\), .test .\31 2u\$\(xlarge\) {
  width: 100%;
  clear: none;
  margin-left: 0;
}
.test .\31 2u\$\(xlarge\) + *,
.test .\31 u\$\(xlarge\) + * {
  clear: left;
}
.test .\-11u\(xlarge\) {
  margin-left: 91.6666666667%;
}
```

The real bug is probably in the selector Parsing here, due to escaped chars in the selectors

Cerdic:
Escaped chars in selectors should not been systematically unescaped in the output, leading to unvalid selectors that the parser itself is refusing as valid. 

For now we are keeping the escaped sequence in the selector, in a normalize way, to not take the risk to produce an unvalid selector 

The target and safer should be to have the same way of managing escaping as in string and keyword: store the real char in a non escaped form in the selector representation, and escape in the output what should be escaped only.

But this require a better structure for representation of selectors out of the parser, otherwise this would more or less implie to reparse in the compiler to know what to escape, which would be a shame
(ie for instance escape non alpha first chars of identifiers and non allowed chars in identifier, but without escaping attributes content matching)

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -15,6 +15,7 @@ namespace ScssPhp\ScssPhp;
 use ScssPhp\ScssPhp\Base\Range;
 use ScssPhp\ScssPhp\Compiler\Environment;
 use ScssPhp\ScssPhp\Exception\CompilerException;
+use ScssPhp\ScssPhp\Exception\ParserException;
 use ScssPhp\ScssPhp\Exception\SassScriptException;
 use ScssPhp\ScssPhp\Formatter\Compressed;
 use ScssPhp\ScssPhp\Formatter\Expanded;
@@ -1809,7 +1810,7 @@ class Compiler
 
             try {
                 $isValid = $parser->parseSelector($buffer, $newSelectors, true);
-            } catch (\Exception $e) {
+            } catch (ParserException $e) {
                 throw $this->error($e->getMessage());
             }
 
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -1807,7 +1807,13 @@ class Compiler
             $buffer    = $this->collapseSelectors($selectors);
             $parser    = $this->parserFactory(__METHOD__);
 
-            if ($parser->parseSelector($buffer, $newSelectors, true)) {
+            try {
+                $isValid = $parser->parseSelector($buffer, $newSelectors, true);
+            } catch (\Exception $e) {
+                throw $this->error($e->getMessage());
+            }
+
+            if ($isValid) {
                 $selectors = array_map([$this, 'evalSelector'], $newSelectors);
             }
         }
--- a/tests/inputs/selectors.scss
+++ b/tests/inputs/selectors.scss
@@ -303,3 +303,13 @@ $end: ' ';
 #{$selector} a .tocnumber #{$end} {
         min-width: 100%;
 }
+
+
+// escaped
+.\31u\$ {clear: left;}
+.\31 u\$ {clear: left;}
+.\31u\24 {  clear: right;}
+.\31 u\24 {  clear: right;}
+.a\31 u\24 {  clear: both;}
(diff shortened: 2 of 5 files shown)


tests/inputs/selectors.scss
```diff
@@ -303,3 +303,13 @@ $end: ' ';
...
+
+
+// escaped
+.\31u\$ {clear: left;}
+.\31 u\$ {clear: left;}
+.\31u\24 {  clear: right;}
+.\31 u\24 {  clear: right;}
+.a\31 u\24 {  clear: both;}
+$x: '\\28' + 'xlarge' + '\\29';
+.\31 2u#{$x}, .\31 2u\24#{$x} { width: 100%; clear: none; margin-left: 0; }
```
stof, line 315: I suggest that we also contribute upstream to have tests covering that in sass-spec, instead of covering them only in our own testsuite (see #179)

src/Compiler.php
```diff
@@ -1807,7 +1807,13 @@ protected function evalSelectors($selectors)
             $buffer    = $this->collapseSelectors($selectors);
             $parser    = $this->parserFactory(__METHOD__);
 
-            if ($parser->parseSelector($buffer, $newSelectors, true)) {
+            try {
+                $isValid = $parser->parseSelector($buffer, $newSelectors, true);
+            } catch (\Exception $e) {
```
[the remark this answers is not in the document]
Cerdic: Fixed

tests/inputs/selectors.scss
Cerdic: Done with https://github.com/sass/sass-spec/pull/1600
Let's see how the processus is working.

Cerdic:
I made the corrections, and I had to correct the patch to be fully compliant on the draft spec I sent as a PR on sass-spec , as dart-sass unescape numbers if not at the first char.

Again, the fix is not a definitive fix, and I'm quite sure that pushing a bit on combination of escaped sequence and position of the escape sequence you could find other cases where scssphp will keep the escaped sequence and not dart-sass.

But at lease, this is a safe fix as the final selector should always be as valid as the input selector - If I did no mistake.
To be followed when we improved the selector parsing and abstract representation

stof:
keeping more escape sequences is OK, if the output is equivalent CSS

tests/inputs/selectors.scss
stof: @Cerdic any plan to take their review into account ?

PR #276, opened 2020-11-17
Fix sourcemaps for the compressed output

stof:
When writing new output to the same line that the previous output, the current column needs to be incremented rather than reset.
For the expanded output, the code was fine as the current column is almost always 0 at that point (due to the resetting happening when writing a new line). But the compressed output is writing everything as a single line.

Closes #275

PR #277, opened 2020-11-17
Improve the pscss CLI

### LINKED ISSUE #267: pscss should support outputing to a file
Currently, `bin/pscss` always writes the CSS to stdout. To be consistent with the CLI of official sass implementation, we should probably also support it in our CLI.
Outputting to a file opens the door to supporting source map files, rather than we currently only support embedded source maps in the CLI.

stof:
- add support for writing the output to a file rather than stdout. Closes #267
- add support for choosing between a source map file or an inline source map when writing to a file (defaults to a map file for consistency with the official sass CLI)
- add support for embedding sources in source maps

### DIFF
--- a/bin/pscss
+++ b/bin/pscss
@@ -32,6 +32,8 @@ $inputFile = null;
 $changeDir = false;
 $encoding = false;
 $sourceMap = false;
+$embedSources = false;
+$embedSourceMap = false;
 
 /**
  * Parse argument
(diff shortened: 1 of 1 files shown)


stof:
I wrote a hacky support for writing a map file when working on #232 because https://sokra.github.io/source-map-visualization/ does not support visualizing inline source maps (and it is a helpful tool to debug the generated source map).
Then, I needed something like that again to work on #276 today. But this time, I decided to implement it in a clean way that can be merged.

The naming of options `--embed-source-map` and `--embed-sources` correspond to the naming used by the official dart-sass CLI.

PR #287, opened 2020-12-30
treat css custom properties as valid css element

### LINKED ISSUE #286: Compiling plain css file imported into SCSS fails when using custom properties
Edited stacktrace:

```
Fatal error:  Uncaught ScssPhp\ScssPhp\Exception\ParserException: Error : `--page-background:#6badff;` isn't
allowed in plain CSS (custom): failed at `--content-background:white;` ../template/public_html/color.css on
line 4, at column 2 in /vendor/scssphp/scssphp/src/Parser.php:205
Stack trace:
#0 /vendor/scssphp/scssphp/src/Parser.php(1229): ScssPhp\ScssPhp\Parser->parseError()
#1 /vendor/scssphp/scssphp/src/Parser.php(1628): ScssPhp\ScssPhp\Parser->assertPlainCssValid()
```

From messing around in the Parser class, I can see that the parsed row has already been correctly identified as a custom property so I believe the problem is that the `isPlainCssValidElement` function has no mention of `Type::T_CUSTOM_PROPERTY`.

I added it to the switch statement locally and the file appears to compile as it does if I rename the files from `.css` to `.scss`.

i.e. here
https://github.com/scssphp/scssphp/blob/c633b96209c47670bd361aaddf2a6a6965278b13/src/Parser.php#L1276-L1282

Please can you confirm if that is the correct approach, worthy of a PR, or if an alternative solution is more appropriate.

willpower232:
resolves #286

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -1366,6 +1366,7 @@ class Parser
                     ]
                 ];
 
+            case Type::T_CUSTOM_PROPERTY:
             case Type::T_UNARY:
                 $parsed[2] = $this->isPlainCssValidElement($parsed[2]);
                 if (! $parsed[2]) {
--- a/tests/specs/sass-spec-exclude.txt
+++ b/tests/specs/sass-spec-exclude.txt
@@ -748,10 +748,6 @@ css/media/range/error/mismatched_range/lte_gte
 css/media/range/static
 css/media/range/with_expressions
 css/moz_document/multi_function
-css/plain/custom_properties/arbitrary_tokens
-css/plain/custom_properties/color
-css/plain/custom_properties/identifier
-css/plain/custom_properties/nested
 css/plain/import/sass_takes_precedence
 css/selector/attribute/modifier/after_string
 css/selector/error/attribute/modifier/digit


willpower232:
@Cerdic it seems to be failing against the scss spec here https://github.com/sass/sass-spec/blob/d975d33146fb679a6b359ceca329012f02e4a794/spec/css/plain/error/statement.hrx#L427-L443

The test seems to be for interpolation so previous test passes were technically false positives because custom properties are allowed, they just can't be based on other values(?)

Should I add another check to Parser to verify the value doesn't contain `{` and `}`?

stof:
The value of the custom property indeed needs to be validated to ensure it is plain CSS, as done for many other types.

willpower232:
Thanks @stof, I had a quick look at `$parsed` and it seems the interpolation was already identified so I added a check directly for it.

<details><summary>parsed example</summary>
```
Array
(
    [0] => custom
    [1] => Array
        (
            [0] => string
            [1] => 
            [2] => Array
                (
                    [0] => --
                    [1] => b
                )

        )

    [2] => Array
        (
            [0] => string
            [1] => 
            [2] => Array
                (
                    [0] =>  
                    [1] => Array
                        (
                            [0] => interpolate
                            [1] => Array
                                (
                                    [0] => keyword
                                    [1] => c
                                )

                            [2] =>  
                            [3] => 
                        )
                )
        )
)
```
</details>

It has passed the tests, I guess the only bit I don't like is how many array keys are manually specified but I guess if the structure was ever going to be changed it would be a massive job anyway :sweat_smile:

src/Parser.php
```diff
@@ -1281,6 +1281,13 @@ protected function isPlainCssValidElement($parsed, $allowExpression = false)
             case Type::T_MEDIA:
                 return $parsed;
 
+
+            case Type::T_CUSTOM_PROPERTY:
+                if ($parsed[2][2][1][0] == Type::T_INTERPOLATE) {
```
stof, line 1286: Please use `$this->isPlainCssValidElement($parsed[2])` instead of hardcoding a particular structure (which might not be valid depending on how it looks like).

stof (changes requested):

src/Parser.php
stof: also, this detects only one kind of SCSS syntax.

willpower232: ah thanks, that makes much more sense

stof (approved):

stof:
And this change makes us pass some tests of the official sass-spec testsuite, which is great. I updated the PR to enable these tests.

Issue #266, opened 2020-11-10
Extract SassScript functions outside the Compiler class


stof:
## Motivation

Currently, the `Compiler` class is huge (approx 9kLOC) which makes it harder to work on them:
- identifying the public API is hard
- such big file slows down IDE (my PHPStorm regularly gets stuck at analyzing the code while I type when working in this file, which then prevents using autocompletion)

The current architecture also suffers from some weaknesses:
- the discovery of available functions and their prototype is magic
- all these `lib*` methods are considered unused by static analysis, because it cannot recognize the way they are used
- overloaded functions (functions with multiple prototypes) have to guess the prototype being used based on the arguments, while the calling code already knows exactly which overload was used
- this architecture would not play well with implementing modules

## Proposal

SassScript functions are moved into a bunch of separate (internal) classes in the `ScssPhp\ScssPhp\Functions` namespace. These functions are grouped by Sass module to which they related (so this namespace would contain `Color`, `Math`, `List`, `Map`, `String`, `Selector` and `Meta` classes, with a suffix to decide upon to avoid issues as we have 2 reserved keywords in that list of names). Functions will be implemented as public static methods in these classes. Overloads will be able to use separate callables to simplify the implementation. Helper functions will become private static methods in the classes using them, or be moved to a separate Util or to the Value API if they need to be reused across modules.

Some `meta` functions will still be implemented in the compiler layer because they require access to internals of the evaluation environment (that's what they are meant to expose). For reference, dart-sass also treat them in a special way for that reason. In practice, I think they will use an infrastructure similar to the one of custom functions. Some meta functions can be implemented standalone and will follow the normal pattern though.

Then, for the registration of functions with their signatures, I suggest to have a `FunctionRegistry` class holding them all. Note that this proposal differs from the dart-sass architecture on purpose to be more friendly to the way PHP works. This class is designed to rely on a big static array that can stay in OPCache shared memory rather than relying on lots of objects that need to be instantiated each time we create a compiler.

```php

namespace ScssPhp\ScssPhp\Functions;

class FunctionRegistry
{
    private static $functions = [
        'str-index' => [ 'arguments' => [ 'string', 'subString' ], 'callback' => [ StringFunctions::class, 'index' ] ],
        'rgb' => [ 'overloads' => [
            [ 'arguments' => [ 'channels' ], 'callback' => [ ColorFunctions::class, 'rgbChannels' ] ],
[...]

PR #292, opened 2021-01-05
Introduce a LoggerInterface

### LINKED ISSUE #223: [RFC] Introduce a logger for warnings
dart-sass has a [`Logger` API](https://github.com/sass/dart-sass/blob/master/lib/src/logger.dart) (exposed only to Dart code, not to JS code, as the JS API is intentionally copying the API of node-sass). This API allows customizing the way warnings and debug messages are reported. Their default logger is writing to stderr

I suggest that we introduce such an API in scssphp, replacing `Compiler::setErrorOuput` (sic).

A potential use case would be for a CMS to collect the warnings and report them in some admin UI or another logging location it has, instead of sending them to stderr, which requires looking at the php-fpm logs to see them. In practice, I think that most CMS users relying on this library are never seeing the warnings (which make them useless in that case, and may hurt in the future when the deprecation becomes a removal and they missed all the timeframe where they could have prepared the update)

On a side note, this could also simplify our tests as we could be using a collecting logger instead of having to deal with a stream that we may leak

stof:
The object-oriented interface makes it simpler to customize the reporting of warnings (no need to use a special stream wrapper to keep it realtime). It will also simplify future internal refactorings as object references are easier to share than PHP resources.
As a nice benefit, the stderr stream opened automatically by the compiler is now properly closed when it is not used anymore.

Closes #223

### DIFF
--- a/scss.inc.php
+++ b/scss.inc.php
@@ -26,6 +26,9 @@ if (! class_exists('ScssPhp\ScssPhp\Version', false)) {
     include_once __DIR__ . '/src/Formatter/Expanded.php';
     include_once __DIR__ . '/src/Formatter/Nested.php';
     include_once __DIR__ . '/src/Formatter/OutputBlock.php';
+    include_once __DIR__ . '/src/Logger/LoggerInterface.php';
+    include_once __DIR__ . '/src/Logger/QuietLogger.php';
+    include_once __DIR__ . '/src/Logger/StreamLogger.php';
     include_once __DIR__ . '/src/Node.php';
     include_once __DIR__ . '/src/Node/Number.php';
     include_once __DIR__ . '/src/Parser.php';
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -20,6 +20,8 @@ use ScssPhp\ScssPhp\Exception\SassScriptException;
 use ScssPhp\ScssPhp\Formatter\Compressed;
 use ScssPhp\ScssPhp\Formatter\Expanded;
 use ScssPhp\ScssPhp\Formatter\OutputBlock;
+use ScssPhp\ScssPhp\Logger\LoggerInterface;
+use ScssPhp\ScssPhp\Logger\StreamLogger;
 use ScssPhp\ScssPhp\Node\Number;
 use ScssPhp\ScssPhp\SourceMap\SourceMapGenerator;
 use ScssPhp\ScssPhp\Util\Path;
@@ -245,10 +247,6 @@ class Compiler
      */
     protected $sourceColumn;
     /**
-     * @var resource
-     */
-    protected $stderr;
-    /**
      * @var bool|null
      */
     protected $shouldEvaluate;
@@ -284,6 +282,11 @@ class Compiler
     private $legacyCwdImportPath = true;
 
     /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * Constructor
      *
      * @param array|null $cacheOptions
@@ -297,7 +300,7 @@ class Compiler
             $this->cache = new Cache($cacheOptions);
         }
 
-        $this->stderr = fopen('php://stderr', 'w');
+        $this->logger = new StreamLogger(fopen('php://stderr', 'w'), true);
     }
 
     /**
@@ -322,15 +325,34 @@ class Compiler
     }
 
     /**
+     * Sets an alternative logger.
+     *
+     * Changing the logger in the middle of the compilation is not
+     * supported and will result in an undefined behavior.
+     *
+     * @param LoggerInterface $logger
+     *
+     * @return void
+     */
+    public function setLogger(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
+    }
+
+    /**
      * Set an alternative error output stream, for testing purpose only
      *
      * @param resource $handle
      *
      * @return void
+     *
+     * @deprecated Use {@see setLogger} instead
      */
     public function setErrorOuput($handle)
     {
-        $this->stderr = $handle;
+        @trigger_error('The method "setErrorOuput" is deprecated. Use "setLogger" instead.', E_USER_DEPRECATED);
+
+        $this->logger = new StreamLogger($handle);
     }
 
     /**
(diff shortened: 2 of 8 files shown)


Cerdic:
I can see that `->warn()` calls also provide the location most of the time, in a different text format.

Should we not split informations in the function signature like for instance
`
public function warn($message, $locationFile=null, $locationLine=null, $deprecation = false) {
}
public function debug($message, $locationFile=null, $locationLine=null) {
}
`

Then both function would be in charge of the final message formating and eventualy at the end calling a unique `->output($text)` method that would be the only one needed to be overriden?

Thus maybe an intermediate `OutputLogger` class could be usefull, that StreamLogger would override for the ouput method, and that the lib customers would also override.

Or if we want to make no choice about implementation of location arguments for now and keep everything easy to improve later, should we not have a `->deprecated($message)` method to avoid the optionnal arg on `warn()` and keep basic function calls with just a string arg, that we can extend in the future?

I think both options are valid but that the PR is a bit in the middle...

stof:
Warnings don't only have a concept of location but also of trace. For now, I'd rather keep that out of the logger signature, until we can figure the right type for passing this info between the compiler code and the logger. Adding separate params for the file and line would make it harder to change the type in the future.

Note that this LoggerInterface has 2 places where it affects extension points:

- custom implementations of the Logger (to collect warnings in a CSM plugin for instance).
- custom functions that want to trigger warnings

In a next step, I expect that functions triggering warnings don't directly call the logger anyway but go through some API that we provide in the Compiler. This "helper" API will be the one responsible for injecting the location and or the trace in warnings. As part of the work on #266, we will actually have a static API for such `warn` calls, that would forward the call to a callable managed by the compiler (and accessing the logger of that compiler). That would work the same than how dart-sass exposes a `warn` API.

I would say that at the logger level, a single `warn` function with a boolean parameter for deprecations is easier to deal with (and we can have a signature `warn(string $message, bool $deprecation = false, ?Location $location = null, ?Trace $trace = null)` in the future in a backward compatible way.

Regarding the SassScript API (used by function implemented outside the compiler, which will include core functions as well in the future), we might separate methods for normal warnings and deprecations if we want. That would avoid having to deal with the optional `$deprecation` parameter when you want to log a normal warning with extra args we might add in the future once we have a better tracking of source locations.
And the trace will probably stay managed by the compiled itself anyway and injected in the call to the logger by the warn callback rather than by custom functions.

stof:
For the context, here is how I imagine the migration path for the day we introduce proper objects to manage the source span and the trace:

```php
interface LocationAwareLoggerInterface extends LoggerInterface {
    public function warn($message, $deprecation = false, SourceSpan $span = null, Trace $trace = null);
    
    public function debug($message, SourceSpan $span = null);
}

/** @internal */
final class MigratingLogger implements LocationAwareLoggerInterface
{
    /** @return LocationAwareLoggerInterface */
    public static function wrap(LoggerInterface $logger)
    {
        if ($logger instanceof LocationAwareLoggerInterface) {
            return $logger;
        }

        @trigger_error(sprintf('Not implementing LocationAwareLoggerInterface for the "%s" logger is deprecated.', get_class($logger)), E_USER_DEPRECATED); 

        return new self($logger);
    }

    // Implement warn() and debug() to build a message containing the location (the existing logic we have in the compiler) and forwarding it to the old logger
}

class Compiler
{
    public function setLogger(LoggerInterface $logger)
    {
        $this->logger = MigratingLogger::wrap($logger);
    }
}
```

The compiler would then always work with a LocationAwareLoggerInterface in places logging things. And this would be fully BC, as the only caller to the logger will be in our own code (custom functions will go through the wrapper) so we can refactor those.

TimWolla:
Did you intentionally opt not to use the PSR-3 API: https://www.php-fig.org/psr/psr-3/? If you simply wasn't aware of it I recommend to take a look for interoperability.

stof:
@TimWolla This is about a logger API tied to the needs of the scssphp compiler, which is more powerful. The common way to report sass warnings will probably not be to send them to your logging infrastructure (in a CMS, you need to show these warnings to the author of the scss file, not to the sysadmin for instance). And they would be reported with details (well, not so much flexibility in this first version as they only get a message, but see the discussion for future plans once we have a proper representation of source locations and traces).

If your own use case is to send such logs to your logging infrastructure, implementing the LoggerInterface on top of a PSR-3 logger would take approximately 10 LOC. But this won't be shipped in the core as that's not the intended way to report sass warnings.

TimWolla:
@stof The PSR-3 interface is not limited to logging to logging infrastructure. It would also allow for a logger that simply stores logged messages in an internal array that can be read out later and then shown in the CMS administrator's dashboard.

By leveraging the `$context` parameter the PSR-3 interface should be no less powerful than what you plan to implement.

> If your own use case is to send such logs to your logging infrastructure, implementing the LoggerInterface on top of a PSR-3 logger would take approximately 10 LOC.

My 2ct would suggest scssphp taking a PSR-3 logger by default, possibly implementing an helper function that properly fills in the `$context` based of line numbers / source location given as the helper's parameters.

stof:
@TimWolla but the *API* of PSR-3 is geared toward the needs of logging infrastructure. Exposing the sass logs through a PSR-3 API would make things harder to discover for integrators (as they would not get a clear contract about which info we give them).

PSR-3 is meant to provide interoperability for generic logger needs. This does not mean that libraries cannot have specialized interfaces to hook into them (this interface could be named `SassWarningReporterInterface` and nobody would even suggest for it to use PSR-3 directly, but the current name is kept similar to the name used in dart-sass for the similar concept in order to make our maintenance easier).
Especially once we can have the v2 of the API, a specialized API unlocks better reporting for projects that want to do so, which would not be easy if they have to reverse-engineer the way data gets fit in a PSR-3 context.

So yeah, that's indeed a very explicit choice not to use PSR-3 but a specialized API here.

Cerdic:
> Regarding the SassScript API (used by function implemented outside the compiler, which will include core functions as well in the future), we might separate methods for normal warnings and deprecations if we want. That would avoid having to deal with the optional $deprecation parameter when you want to log a normal warning with extra args we might add in the future once we have a better tracking of source locations.

I would have a preference for a separated `deprecated()` method then to avoid the optional `$deprecation` arg on `warn()` but this is not a blocking point for me

Cerdic (approved):

stof:
@Cerdic I also have the preference for the higher-level API that will be used in SassScript implementations. But that's not implemented in this PR. This will be the next one (a prerequisite for allowing to extract functions btw).
For the LoggerInterface, I think things are actually easier with a single method rather than 2.

PR #293, opened 2021-01-20
Register an autoloader in the scss.inc.php file

stof:
This makes it a lot easier to maintain than using a list of require statements, especially regarding the order of loading files once we have more interfaces or inheritance involved. As our package source code follows PSR-4, the autoloader implementation can support any future class we add.


stof:
The only impact for projects using this file would be for projects relying on the implicit `__autoload` autoloader function instead of using `spl_autoload_register`. I consider this a non-issue:
- `__autoload` is deprecated since PHP 7.1 and unsupported in PHP 8.0
- such projects are already incompatible with any other code using `spl_autoload_register`
- such projects can fix that easily by registering their `__autoload` function explicitly in `spl_autoload_register`

This also changes what happens for `class_exists(SomeScssPhpClass::class, false)` after loading the inc file (as the class won't be loaded directly). But code wanting to detect the availability of a class should be using ``class_exists(SomeScssPhpClass::class)` anyway to support autoloading features (when using the library through Composer, it is already using autoloading rather than early require calls)

Issue #265, opened 2020-11-09
Values as typed objects


stof:
This issue serves as a detailed spec for the refactoring of the representation of values for #193 

## Design of value objects

Value objects are immutable. Their API is modelled after the dart-sass classes for values, with a few differences:

- we don't distinguish between an internal implementation and an external type. This is a "hack" used in dart-sass to workaround the absence of compiler-enforced internal APIs  in dart (which will be solved in the next version of dart AFAIK so dart-sass will probably change that in the future anyway). In the PHP ecosystem, IDEs are warning when using methods tagged as `@internal`. Psalm also reports that as an error during static analysis (for phpstan, there is [a feature request](https://github.com/phpstan/phpstan/issues/1178)). So using `@internal` will be considered good enough
- some methods will have slightly different names due to PHP reserved keywords being forbidden in method names on PHP 5.6 (for instance `SassList.empty()` from dart will become `SassList::createEmpty()` in PHP)
- overriding the `==` operator is not possible, so we'll implement a `equals(Value $other): bool` method that should be used. that method should be used in any code wanting to compare 2 Sass values
- methods returning a `List<Value>` will return a PHP array (which solves easily the fact that dart enforces that returned collections are not modifiable, as PHP array are not passed by reference)
- for `Map<Value, something>`, we'll need to find a replacement API. `SplObjectStorage` is not suited for that as it uses `spl_object_hash` for equality by default, and its extension point `SplObjectStorage::getHash` would be very hard to use to implement equality rules (In other languages, like dart for instance, classes have a `hashcode` API returning an integer. But then, for such equality checks, objects returning the same hashcode are still compared with `==`. So the implementation of `hashcode` can return false positives but not false negatives, which makes things a lot simpler)

## Implementation plan

The goal of this plan is to perform an incremental migration of the codebase. This is probably not meant to be released until the end of the migration though, to avoid exposing the migration layer in releases (and then having to keep it for backward compatibility). But that will allow to keep a working 2.0-dev compiler during the migration (and spreading the migration across separate commits/PRs to make it easier).

1. Implement `Value` and its subclasses
2. Add support for `Value` in the compiler (not reading `$value[0]` on them, making them no-op in `reduce` and maybe a few other stuff)
3. ~~Add support for  `Value` in the formatter (being able to output them)~~ currently, this is handled in `Compiler::compileValue` actually, the formatter does not deal with our array-based value API but a separate API.
[...]

PR #301, opened 2021-03-04
Add an object-oriented representation of sass values

stof:
Refs #265 

TODOs:
- [x] Implement `SassMap` properly
- [x] add tests for the Value APIs
- [x] ~~Implement color math (which is not there in dart-sass, but is only deprecated on our side)~~
- [x] Fix the string serialization for PHP < 7.4 (thanks CI)
- [x] Update the serialization implementation to account for updates in dart-sass for the string serialization
- [x] Implement `SassCalculation::equals()`

Using this API in our compiler will be done in follow-up PRs.

### DIFF
--- a/src/Collection/Map.php
+++ b/src/Collection/Map.php
@@ -1,0 +1,163 @@
+<?php
+
+/**
+ * SCSSPHP
+ *
+ * @copyright 2012-2020 Leaf Corcoran
+ *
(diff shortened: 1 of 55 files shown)


stof:
The implementation of those value objects is now complete. I'm merging it to let me use it in follow-up PRs (note that this PR alone does not bring any value to consumers of the library yet, as those objects are not yet integrated with the compiler)

PR #306, opened 2021-03-09
Implement static analysis with phpstan

stof:
phpstan is installed using bamarni/composer-bin-plugin so that it does not impact the dependency resolution of the root dependencies. This is necessary because phpstan requires PHP 7.1+.

Due to all our legacy array-based code for values and AST, phpstan complains about a lot of things that are not properly typed. The baseline is used to ignore them. phpstan still helps finding some issues today, and will help a lot when modernizing the codebase.

The process when contributing looks like that:
- make your changes
- run `make phpstan`
- fix as many new issues as possible (this might involve multiple iterations)
- once left only with unfixable errors due to the legacy, run `make phpstan-baseline` to regenerate the baseline


mahagr:
Just be careful when fixing the issues, I've run a few times into code which depends on wrong behavior in order to work, lol.

stof:
@mahagr in our legacy codebase, "fixing the issue" is most of the time about fixing the incorrect phpdoc (look at how many places were missing `|null` in #305 for instance). And I don't expect to have phpstan being green without baseline on the legacy parts of the codebase either. But I expect that the modernized parts of the codebase (see #301 for a first part of that work) to stay out of the baseline.

robocoder:
Nice!

We should add that process to README, or create a CONTRIBUTING doc.

stof:
@robocoder good idea. I updated the readme.

PR #307, opened 2021-03-09
Clean code based on phpstan

stof:


robocoder (approved):
LGTM. I especially like 02b7006d.

PR #312, opened 2021-03-10
Fix the current version on the website

stof:
This was forgotten when doing the last releases


stof:
Note that I'm currently documenting the release process.

PR #313, opened 2021-03-10
Document the release process

stof:


robocoder:
I'd like to drop the v from the tag but otherwise LGTM.

Issue #310, opened 2021-03-10
[RFC] Revised plan for 2.0


stof:
After staring my work on #265, I realised that merging that work into 1.x is too risky regarding backward compatibility, due to #201. The existing array-based value API is less type-capable than the new API (see #296 for instance), so we need to refactor the Environment to use the `Value` objects to store variables. This will break BC for child classes messing with the internals of the Compiler.

## Suggested roadmap

> Note, when I give versions like `1.x`, it means a minor version in the series, with the exact version number depending on which this gets released. Not everything might happen in the same minor release.

### 1.x

- add a deprecation warning when a `lib*` method is detected as the implementation of a function and it comes from a child class (asking to use `registerFunction` instead)
- tag all protected functions as `@final` and `@internal` so that child classes using them or overriding them get a warning in the IDE
- do the same for public methods that should not be needed for implementing custom functions and don't need to stay public (for instance, this does not include the `Compiler::assert*` helpers as custom functions need them).
- mark public helpers as `@final` so that child classes get an IDE warning when overriding them
- tag classes as `@internal` for all classes that should not belong to the public API (this includes the `Parser` class to me)

### 2.0

- remove color math (which is already deprecated since 1.4.0 and is gone entirely in dart-sass)
- remove deprecated APIs of the Compiler
- remove deprecated formatters (this will unlock rewriting the way we produce CSS in a more modern way as we won't have to deal with these old styles)
- change all protected APIs of the Compiler to `private` (as well as public internal methods that don't need to stay public)
- change all `@final` public methods of the Compiler to actually be final

The 2.0 release is intentionally kept small in scope so that it can happen fast after the final 1.x release, to avoid maintaining 2 branches in parallel for a long time.

### 2.x

- implemented typed value objects (#265)
- rewrite the scope handling to store variables as typed values
- update `registerFunction` to support registering a function as using the modern value API (and deprecate registering functions with the old API)
- etc...

Over time in 2.x, we can rewrite parts of the library (see https://github.com/scssphp/scssphp/issues/193) freely, as almost everything will be internal and so won't impact backward compatibility.
And hopefully, after some time, we will have a 2.Y release passing sass-spec.

## Suggested process

- implement the 1.x changes
- release 1.x
- create a `1.x` branch
- change the `master` branch alias to `2.x`
- clean deprecated APIs
- do the 2.0 release
[...]

PR #315, opened 2021-03-11
Rework CompilationResult

stof:
This reworks the CompilationResult result object introduced in #270. This has to be merged before the next release (otherwise CompilationResult will be covered by BC and prevent the changes)

- `Scssphp\Scssphp\CompilationResult` is now an immutable value object instead of a mutable one
- `Scssphp\Scssphp\CompilationResult` now longer exposes some internal API publicly
- metadata used only by the caching layer are not exposed in the CompilationResult anymore. They are kept internal (the object stored in the cache is an internal `CachedResult` object wrapping the CompilationResult)
- metadata collected during the compilation are stored in properties of the Compiler (that are reset at the end of the compilation to free memory, except for `parsedFiles` which has a deprecated getter)
- the Compiler no longer retains a reference to the CompilationResult after returning it (which kept reference to much more memory than just `parsedFiles` as there is the CSS and sourcemap in there)
- `getParsedFiles` is not exposed in CompilationResult, in favor of `getIncludedFiles`, which returns a list of file paths. When looking at various open-source packages using scssphp, any usage of `$compiler->getParsedFiles()` I found were actually usages of `array_keys($compiler->getParsedFiles()`

/cc @Cerdic as you implement the initial version.

### DIFF
--- a/src/CompilationResult.php
+++ b/src/CompilationResult.php
@@ -40,7 +40,6 @@ class CompilationResult
      */
     private $sourceMapUrl;
 
-
     /**
      * All the effective parsed files
      * @var array<string, int>
@@ -172,21 +171,10 @@ class CompilationResult
         return $this->includedFiles;
     }
 
-    // A map from source file URLs to the corresponding [SourceFile]s.
-    //
-    // This can be passed to [sourceMap]'s [Mapping.spanFor] method. It's `null`
-    // if source mapping was disabled for this compilation.
-    public function getSourceFiles()
-    {
-
-    }
-
-
     public function __toString()
     {
         return $this->getCss(); // To reduce the impact of the BC break
     }
-
 
     /**
      * Store the sourceMap and its storage data
--- a/src/CompilationResult.php
+++ b/src/CompilationResult.php
@@ -209,7 +209,7 @@ class CompilationResult
      * The sourceMap Css content, if there is a sourceMap
      * @return string
      */
-    public function getSourceMapCss()
+    private function getSourceMapCss()
     {
         if ($this->sourceMap) {
             if ($this->sourceMapFile) {
--- a/src/CompilationResult.php
+++ b/src/CompilationResult.php
@@ -47,12 +47,6 @@ class CompilationResult
     private $parsedFiles = [];
 
     /**
-     * All the @import files and urls seen in the compilation process
-     * @var string[]
-     */
-    private $includedFiles = [];
-
-    /**
      * All the @import files resolved and imported (use to check the once condition)
      * @var array
      * @phpstan-var list<array{currentDir: string, path: string, filePath: string}>
@@ -116,7 +110,6 @@ class CompilationResult
     public function addImportedFile($currentDirectory, $path, $filePath)
     {
         $this->importedFiles[] = ['currentDir' => $currentDirectory, 'path' => $path, 'filePath' => $filePath];
-        $this->addIncludedFile($filePath);
     }
 
 
@@ -142,33 +135,11 @@ class CompilationResult
     }
 
     /**
-     * Save the included files
-     * @param string $path
-     *
-     * @return void
-     */
-    public function addIncludedFile($path)
-    {
-        // unquote the included path if needed
-        foreach (['"', '"'] as $quote) {
-            if (strpos($path, $quote) === 0 && substr($path, -1) === $quote) {
-                $path = substr($path, 1, -1);
-                break;
-            }
-        }
-
-        $this->includedFiles[] = $path;
-    }
-
-    /**
-     * For filesystem imports, this contains the import path. For all other
-     * imports, it contains the URL passed to the `@import`.
-     *
      * @return string[]
      */
     public function getIncludedFiles()
     {
-        return $this->includedFiles;
+        return array_column($this->importedFiles, 'filePath');
     }
 
     public function __toString()
--- a/src/CompilationResult.php
+++ b/src/CompilationResult.php
@@ -21,11 +21,6 @@ use ScssPhp\ScssPhp\Exception\CompilerException;
 class CompilationResult
 {
     /**
-     * @var bool
-     */
-    private $isCached = false;
-
-    /**
      * @var string
      */
     private $css = '';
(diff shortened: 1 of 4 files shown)


Cerdic:
I think I'm a bit lost.

The idea of the CompilationResult was (for me) exactly to group *and* expose all informations needed to manage cache and/or cache invalidation logic to avoid in the future any new problem about consumers having issue with the cache invalidation : if the default (improved) logic was not fitting their need it was possible to manage it your own way.

That said, I can fully see your logic of trying to lower as much as possible the public and exposed side of the lib, to allow a easier evolution in the future. 

But here I have the feeling we're almost back to the start, the CompilationResult beeing just a container for some additional data to the plain css, and not giving any more information than before (ie the fact that the result is a cached result or not, and the list of resolvedImports which is the keypoint if you want to make your own check on invalidation mecanism)

I have to say I'm not available enough to spend sufficient time and dig on this subject to have a fully clear view.

src/CompilationResult.php
```diff
@@ -71,101 +65,19 @@ public function getCss()
...
-        return $this->importedFiles;
-    }
-
     /**
      * @return string[]
      */
     public function getIncludedFiles()
     {
-        return array_column($this->importedFiles, 'filePath');
+        return $this->getIncludedFiles();
```
Cerdic, line 73: Mistake here I think. Should be `$this->includedFiles;`

stof: indeed

stof:
@Cerdic *using* the resolved import actually relies on being able to use `Compiler::findImport`, which is pretty much something I'd like to avoid having as part of the public API of the Compiler in 2.0. That's my reasoning for keeping it out of the public API.
Thus, the structure of this `resolvedImports` array is tied to the *current* implementation of the import resolution algorithm. But I know for sure that we would have to change it when we implement modules (the resolution rules for `@import` and `@use` have some differences due to the import fallback for migration paths). Making this structure public would prevent implementing modules.

#270 claimed to solve 2 issues at once:
- #222, i.e. not having a stateful `Compiler::getParsedFiles` API in favor of accessing the list of file from a result object. This is still the case in this PR (except that I modelled the API after the original suggestion in #222 by returning a list of included files, instead of requiring to extract them from array keys)
- #268 about cache not being invalidated when import resolution changes. This PR still solves that, but does not expose the associated data publicly (caching invalidation logic is not extendable anyway as it happens internally)

> the fact that the result is a cached result or not

why should you care whether the Compiler was able to reuse its cache or had to do the compilation again ? You still get the same result in both cases. I don't see a need for that.

> But here I have the feeling we're almost back to the start, the CompilationResult beeing just a container for some additional data to the plain css

That was **precisely** the request in #222: being a container against the extra additional data that was accessible through a stateful API of the Compiler before (which is an awful way to access it as the usability of `Compiler::getParsedFiles` depends on the previous usage of the compiler and the freeing of memory depends on the next usage)

> The idea of the CompilationResult was (for me) exactly to group _and_ expose all informations needed to manage cache and/or cache invalidation logic to avoid in the future any new problem about consumers having issue with the cache invalidation : if the default (improved) logic was not fitting their need it was possible to manage it your own way.

AFAICT, your implementation still does not allow any hooking inside the cache invalidation. You only get your hand on the CompilationResult object *after* the Compiler is done with it, and so caching is already done by that time.

If the goal is to allow disabling the built-in compiler cache to handle caching in the caller while still having a proper cache invalidation, I can suggest an alternative API. We could have `CompilationResult::getCacheInfo(): CacheInfo` (name to determined, I'm not satisfied with that one) returning an internal object, and `Compiler::isFreshCache(CacheInfo $cacheInfo, $checkImports = true): bool` checking whether the cache is still fresh. **Everything** inside this `CacheInfo` object would be internal. Its only public API would be that you can pass it to `Compiler::isFreshCache` to get freshness info (and that you can serialize it for storage, so we would not put any Closure in it for instance). That way, we can do any refactoring we want in the cache invalidation logic without BC breaks (as the metadata would be internal).
That would basically replace my `CachedResult` wrapper with an object available in `CompilationResult`

Cerdic:
Feel free to go, as I said my mind is not able to get the big picture right now and this is not helping to challenge you

stof:
@Cerdic the fact that you want to be able to check for cache invalidation in the embedder (see #316) is actually a good challenge for that suggested API. But do you need the embedded to be able to use the data of `resolvedImports` by themselves or would it be OK for you to have the blackbox implementation that I suggested in my previous comment (that allows us to make the logic itself evolve) ?

Cerdic:
It depends of the level of consistency you need and if you can deal with some edge case. I was dealing with the previous invalidation implementation relying only on the `importedFiles` but this was an issue for some users.

If you want a reliable invalidation, the acces to the `resolvedImports` data is necessary, but as you said you also need to access to the `resolveImport()` method. 

I think the implementation of the invalidation check in scssphp is now reliable enough to avoid the need of this api for now and we could postpone this work as there is already a lot to do besides that - except if we decide to get rid of the cache mecanism in the lib.

mahagr:
My biggest issue with CompilationResult is that it saves the source map (if not embedded) into the filesystem. What if I wanted to upload the file into CDN instead of saving it locally? Or use FlySystem to save both the CSS and map?

The class should only do what it is supposed to.

stof:
@mahagr saving the sourcemap in the filesystem is indeed the point that still bothers me. But as long as we keep returning it from `compile()` instead of creating a new API (see my comment in https://github.com/scssphp/scssphp/issues/310#issuecomment-796978976), we cannot avoid it to avoid too much BC breaks.

@Cerdic for now, I think I will merge this PR without the external access to the cache invalidation logic. Exposing that publicly later (if requested) is easy to do in a backward compatible way (while removing it from the public API is not)

Cerdic:
@stof let's go for that

mahagr:
@stof Can you at least provide the alternate method to return CSS and map separately without saving anything to a file? There must be some way to do that without breaking things. :)

stof:
@mahagr see #320 where I discuss the remaining points about that API.

PR #318, opened 2021-03-11
Rework the Framework tests

stof:
- properly compile the Foundation styles instead of only parsing the definition of mixins, to ensure better compat
- don't catch exceptions before asserting they are null. Letting PHPUnit catch the exception allows a much better reporting in case of failure.

### DIFF
--- a/tests/FrameworkTest.php
+++ b/tests/FrameworkTest.php
@@ -14,48 +14,39 @@ namespace ScssPhp\ScssPhp\Tests;
 
 use PHPUnit\Framework\TestCase;
 use ScssPhp\ScssPhp\Compiler;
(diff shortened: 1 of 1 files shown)


stof:
I discovered the issue with Foundation when profiling the compilation of both Bootstrap and Foundation. For Foundation, I got a profile where 80% of the time was spent in the Parser (quite unexpected when the result for Bootstrap was 80% spent in the Compiler). Looking at the output of the compilation, I discovered that the generated CSS had only the license comment, because Foundation only defines mixins (and this explains the profile, as the Compiler was not compiling any styles)

stof:
I also used a QuietLogger to avoid sending potential Sass warnings to stderr while running tests.

Issue #209, opened 2020-10-29
The precedence to resolve function definitions is not consistent with dart-sass


stof:
In dart-sass the resolution of functions follows this order of preferences:
- user defined callables (defined in sass using `@function`) (and any namespaced functions coming from imported modules)
- core global builtin callables (implemented in dart-sass itself)
- caller-provided global builtin callables (passed to the `compile()` function and so implemented in Dart or JS depending on the build being used)
- plain css callable (not for `get-function` obviously, only for evaluation function expressions in the AST)

In scssphp, we currently resolve them in this order:
- sass functions (defined using `@function`)
- custom functions registered through `Compiler::registerFunction` (which we call `user functions`, so be careful when comparing implementations)
- custom functions implemented as `ChildCompiler::libMyFunction` (as we don't forbid extending the compiler and overriding existing implementations of functions)
- core functions (implemented as `Compiler::libMyFunction`)
- plain css callable

Basically, what this means is that we allow other PHP code to replace the core functions, while dart-sass allows adding new functions but not replacing the core ones. I think the dart-sass behavior is better, as it ensures that core functions have the expected behavior, making the sass files self-contained.

I suggest deprecating the possibility to override core functions:

- when using `registerFunction`, we would check if the name conflicts and trigger a deprecation
- all `lib*` methods (and their associated `$lib*` prototype static property) would be tagged as `@final` so that IDEs and other tooling (for instance the `DebugClassLoader` from `symfony/error-handler`) can report overrides as being issues
- optionally trigger deprecations when instantiating an inherited Compiler class if it overrides such methods. I'm saying optionally because it would have some performance cost to run the check (no cost if the class is not extended though)

As an alternative to the detection of inherited methods and the tagging of all existing functions, we could deprecate extending `Compiler` altogether (with a `@final` at the class level and a simple check in the constructor to throw a deprecation warning). That's a better idea on the long term anyway, as it would clearly define the public API covered by semver (see #201)

mahagr:
We are overriding some of the library methods to add support for php streams. It allows us to use urls like: `gantry-theme://images/my.jpg` and resolve the relative path in WP, Joomla and Grav CMS -- which have all very different paths to the files.

https://github.com/gantry/gantry5/blob/feature/v5.5.0/src/classes/Gantry/Component/Stylesheet/Scss/Compiler.php#L111

Same is true for imports, we have some custom logic to add proper lookup paths as we support overrides (lib -> theme) over overrides (theme -> user custom):

[...]

PR #319, opened 2021-03-11
Deprecate registering custom function through inheritance

stof:
Such cases now trigger a deprecation warning hinting towards using `registerFunction` instead (when it is a new function and not an override of a core one).


stof:
The deprecation warning is triggered on first usage of the function (to avoid having to check all methods when instantiating the compiler).

stof:
this implements the deprecation warning for half of #209

PR #325, opened 2021-03-13
Deprecate the scssphp-glob function

### LINKED ISSUE #219: [RFC] Deprecate the scssphp-glob() function
In response to https://github.com/scssphp/scssphp/issues/88, a `scssphp-glob()` function was implemented in scssphp.

To me, this function does not belong to the core (we should not provide any additional features specific to scssphp IMO). It should be registered as a custom function using `registerFunction` (which is totally usable to implement such feature today).
The root cause of #88 is that Contao does not give access to our `registerFunction` method. But that should be solved by requesting the feature to Contao instead, not by making this a scssphp built-in function.

stof:
This function is not standard in Sass. It is totally untested. And it has a confusing behavior: the argument must either be a glob based on an absolute path (not easy to do in a Sass file if it must be portable) or relative to the current working directory of the PHP process, not to the Sass file using the function, (which is not in the control of the author of the Sass file).

Closes #219

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -9554,6 +9554,10 @@ class Compiler
     protected static $libScssphpGlob = ['pattern'];
     protected function libScssphpGlob($args)
     {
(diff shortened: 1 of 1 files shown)


stof:
I decided to report the deprecation both to the PHP side (which can provide a replacement) and to the Sass author.

PR #329, opened 2021-03-13
Add CI jobs running on Windows and macOS

stof:


stof:
Here we go. Our testsuite is now compatible with Windows. 2 fixes were needed:
- when writing tests for the `Path::join` utility, I missed some cases needing to account for windows directory separator.
- our own testsuite was misusing the `setVariables` API and was working only by luck (I think it might have broken on other operating systems as well depending on the path in which the repo is cloned, but I haven't tried to use a path including spaces to confirm that)

This will make it easier for contributors if any of them uses Windows.

PR #330, opened 2021-03-13
Deprecate --dump-tree in the CLI

### LINKED ISSUE #317: [RFC] Deprecate the --dump-tree option of the CLI
the `pscss` CLI supports a `--dump-tree` flag (with a `-T` shortcut) which will parse the input file and then dumps the result.

To me, this is not really useful (nor usable):

- the output of the Parser is a mix of array-based AST (for expressions) and `Block` objects. But `--dump-tree` converts everything to associative arrays before dumping. Thus the output uses `print_r`, which is not precise about types. So this cannot actually be used to investigate what needs to be supported. this means that this flag is not really usable for contributors. Adding `var_dump` calls in the Parser or using a debugger is necessary to get useful insights.
- we make no guarantee about the structure of this AST (and the output of èprint_r` is not meant to be machine-readable). So this flag is useless for users of the library

stof:
Closes #317

### DIFF
--- a/bin/pscss
+++ b/bin/pscss
@@ -76,7 +76,7 @@ Options include:
     --help              Show this message [-h, -?]
     --continue-on-error [deprecated] Ignored
     --debug-info        [deprecated] Ignored [-g]
-    --dump-tree         Dump formatted parse tree [-T]
+    --dump-tree         [deprecated] Dump formatted parse tree [-T]
     --iso8859-1         Use iso8859-1 encoding instead of default utf-8
     --line-numbers      [deprecated] Ignored [--line-comments]
     --load-path=PATH    Set import path [-I]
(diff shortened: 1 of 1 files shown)


bin/pscss
```diff
@@ -180,6 +180,8 @@ if ($dumpTree) {
 
     print_r(json_decode(json_encode($parser->parse($data)), true));
 
+    fwrite(STDERR, 'Warning: the --dump-tree option is deprecated. Use proper debugging tools instead.');
+
     exit();
```
stof, line 185: should I change that to `exit(1)` to report it as an error to use it ?

stof: @robocoder @Cerdic what do you think about this ?

stof: thinking about that again, changing the exit code does not make sense. It is only a warning, not an error.

PR #333, opened 2021-03-13
Deprecate the Compiler::setEncoding method

### LINKED ISSUE #322: [RFC] Deprecate Compiler::setEncoding
scssphp currently has partial support for encoding.
Setting anything else than `utf-8` (case insensitive) or a falsy value (for instance the `null` default) in `Compiler::setEncoding` will make the Parser remove the `/u` modifier from all its regexes. But that's really the only change. The parser still assumes an ASCII-compatible encoding (fortunately, all core Sass features are using ASCII names only, which means that the parser should more or less work fine for ASCII-compatible) everything done in the Compiler when dealing with SassScript string still hardcodes `UTF-8`.
Regarding spec compliance, support for arbitrary encoding does not really make sense IMO. The Sass language spec explicitly defines sass string indexes as being based on Unicode codepoints. Our own implementation changing the parser itself is probably not compliant.

The non-UTF-8 parsing is totally entirely untested. The testsuite never use the `Compiler::setEncoding` method.

## Encoding support in official Sass implementations

dart-sass only supports UTF-8 because `dart:convert` does not support UTF-16 yet but only UTF-8 and Latin1 (and they haven't bothered implementing encoding support and detecting only for Latin1).
libsass requires using UTF-8 encoding. If the file is not valid UTF-8, it fails. As explained in https://github.com/sass/libsass/blob/master/docs/unicode.md, there is no plan to change that, as that would require making libsass dependant on libiconv, which was rejected.

ruby-sass used to support more encoding. However, it was detecting the encoding per parsed string (so potentially differently for each imported file), based on the following heuristic:
- check for a UTF-8, UTF-16LE or UTF-16BE BOM at the beginning of the string
- check for a `@charset` at the beginning of the string (nothing allowed before that), following the CSS spec: https://www.w3.org/TR/css-syntax-3/#determine-the-fallback-encoding
- trust the encoding of the ruby string (Ruby strings have an associated encoding. they are not just raw bytes like in PHP)
- fallback to UTF-8 if the ruby string is associated with a binary encoding
Then, it would convert the string from that encoding to UTF-8 (and throw an error if the conversion fails). all parsing would happen in UTF-8 then. This means that the Sass parsing itself was still implemented only for UTF-8 (unlike our parser)

Note that the Sass spec explicitly defines that the *output* of Sass is always using UTF-8 encoding (there is an opened ticket on dart-sass to add an ascii-only mode where all non-ascii chars are escaped in CSS, but that would still respect the rule of outputting UTF-8 as UTF-8 is ascii-compatible. And it is low priority in dart-sass).

## Usage in projects

- SPIP: not used
- Shopware: not used
- Contao: not used
- Gantry: not used
- October CMS: not used

All these projects are requiring the usage of UTF-8 for the scss sources.

Given that I don't think our encoding support actually works, this is not really surprising.

## Proposals

I see 2 possible ways to implement a compliant state regarding encoding.

Note that both proposals get rid of `Compiler::setEncoding`

### 1. Remove support for non-UTF-8 encoding entirely

As the official implementations of Sass only support UTF-8 (given that ruby-sass is long dead now), this could be considered acceptable.
Any input not being valid UTF-8 would trigger an error (which is *probably* already kind of happening today when you don't use `setEncoding` as our unicode regexes will not work, and so parsing will break. But we would provide a clear error message instead of whatever weird error happens later).

### 2. Add proper support for per-file encoding

When starting parsing, we would guess the encoding (the algorithm corresponds the [CSS spec](https://www.w3.org/TR/css-syntax-3/#determine-the-fallback-encoding)):
- if there is a BOM, base the encoding (UTF-8, UTF-16BE or UTF-16LE) on that
- if the beginning of the string is a ASCII `@charset` rule, use that as the encoding (following the CSS rules, so this cannot detect it as UTF-16)
- otherwise, assume UTF-8

As PHP strings don't have an associated encoding, we cannot do the equivalent of ruby's behavior. And trying to detect the encoding with `mb_detect_encoding` would be too fragile IMO (plus, iconv does not have an equivalent API for the fallback).

Then, we would convert the input to UTF-8 (using `mb_convert_encoding` or `iconv`), and fail if we cannot get valid UTF-8.
The whole parsing logic would run in Unicode mode, which is required to properly parse Sass anyway.

which proposal should we choose ?

stof:
The Sass language parsing is specified in term of Unicode (as is CSS).
The implementation of encoding in the compiler is not about converting from the provided encoding to unicode before parsing, but about replacing the parser with a non-compliant one which is not based on Unicode.

Closes #322

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -5693,12 +5693,18 @@ class Compiler
      *
      * @api
      *
-     * @param string $encoding
+     * @param string|null $encoding
      *
      * @return void
      */
     public function setEncoding($encoding)
     {
+        if (!$encoding || strtolower($encoding) === 'utf-8') {
+            @trigger_error(sprintf('The "%s" method is deprecated.', __METHOD__), E_USER_DEPRECATED);
+        } else {
+            @trigger_error(sprintf('The "%s" method is deprecated. Parsing will only support UTF-8 in ScssPhp 2.0. The non-UTF-8 parsing of ScssPhp 1.x is not spec compliant.', __METHOD__), E_USER_DEPRECATED);
+        }
+
         $this->encoding = $encoding;
     }
 


src/Compiler.php
```diff
@@ -5693,12 +5693,18 @@ private function getPrettyPath($path)
      *
      * @api
      *
-     * @param string $encoding
+     * @param string|null $encoding
```
stof, line 5696: `Compiler::$encoding` accepts `null`, so this was some incomplete phpdoc

Issue #309, opened 2021-03-10
Overhaul the registration of custom variables


stof:
the `Compiler::setVariables` API is currently quite messy in its signature. It accepts an `array<string, mixed>` with the following behavior for the mixed value depending on its type (this behavior actually happens in `Compiler::injectVariables`, as `setVariables` stores them without any transformation):

- try to parse it as a string containing a Sass value (there is a bug today where this is tried even if the value is not a string, which makes everything following it quite hard to use)
- `Scssphp\Scssphp\Node\Number`: use it as is (it is already in our value representation)
- `array`: use it as is, by assuming that it is our value representation
- `boolean`: convert to sass boolean
- `null`: convert to sass null
- numeric value: convert to `Number` (even if the input is a *string* containing a number)
- empty string: the Sass empty string
- `string` matching the color syntax: convert to sass color
- `string` (other): convert to `T_KEYWORD`, i.e. a sass unquoted string

This has several shortcomings:
- trying to parse the value as a string containing a Sass value does not even check that the value is indeed a string, which means that everything else is unlikely to work (passing a Number will *probably* work by casting it to string which gives a CSS representation of the number, but that will produce a different number when having complex units)
- this API is almost impossible to explain, and so to document (and to my knowledge, this issue is the first place which is actually trying to explain it)
- `setVariables` is actually merging the provided variables with existing registered variables, which makes it impossible to reset the list without using `unsetVariable` for each name (which is quite unexpected given its name)
- variable names are stripped from their leading `$` and normalized (dashes and underscores are equivalent in Sass variable names) only when injecting. Combined with the fact that `setVariables` merges the existing array, this can lead to weird behavior where your new variable does not work because there is another key in the array overriding it.
- the coercion of PHP values is incomplete

## Proposed changes

- add a `is_string` check before trying to parse the value as a sass source (avoids weird array-to-string casts or other weird errors as `true` casts to `1` which is then parsed as a number), which actually makes it possible to pass our sass values (`array|Number`) properly
- add a `Compiler::replaceVariables` method which accepts an array without merging it (and so resetting variables can be done with an empty array)
- maybe: add a `Compiler::addVariables` method with the existing behavior of `setVariables`, deprecating `setVariables` in favor of choosing the right method between `addVariables` or `replaceVariables`. This would help projects to discover that they actually want a `replaceVariables` behavior because they expected `setVariables` to work that way.
[...]

Issue #303, opened 2021-03-05
Rethinking the handling of warning in sass-spec


stof:
Currently, our SassSpecTest is always asserting the default `warning`, not considering alternative `warning-dart-sass`. This is not consistent with our handling of the output.

Currently, libsass renders warning without any location info (and we do it as well in most cases for now, which is how we pass warning checks for many cases) while dart-sass adds full error location as for errors. When I tried introducing the higher-level API for the logger, managing the location automatically, it broke tests as we would then not match the libsass warning anymore.

Another thing related to the matching of warnings is which version is mentioned in deprecation warnings. libsass says `in Sass 4.0` while dart-sass says `in Dart sass 2.0.0`. For now, we follow libsass. But `Sass 4.0` is not a thing anymore. The Sass language itself is not versioned anymore. It is a living spec.

Improvements to implement/discuss:
- support alternative warnings for the assertions in SassSpecTest -> done in #464 
- Remove the code snippet extract from the warning output before matching, until we have an equivalent formatting in Scssphp (which will require quite some work), keeping only the message before it and the trace (Note: this might help us move forward matching the error message too btw).
- Change the trace formatting in our errors and warnings to match the dart-sass one rather than the libsass one
- Discussion: which version should we mention in deprecation warnings ? The dart-sass one or our own version (with a rewrite of expected warnings before matching in SassSpecTest) ? I don't think we want to remove a deprecated feature before it is removed from dart-sass, so mentioning our own major version could be weird (we might need to change it if we do our own major version before dart-sass). In any case, I think we should move away from using `Sass 4.0`.

What do you think @Cerdic @robocoder ?

robocoder:
Output multiple version numbers ala User-Agent?

stof:
In the new compiler, I'm using `in Dart Sass 2.0.0` for now, allowing to assert the warning output in sass-spec without having to do special replacements in the runner.

stof:
In 2.0, we have the same code snippets than dart-sass, thanks to the `scssphp/source-span` package, so we now follow the dart-sass output for warnings.

PR #335, opened 2021-03-13
Add a high-level API to report warning

stof:
The goal of this API is to allow implementing custom functions without the need for a reference to the Compiler object. This is the next step after #239 which allowed avoid the Compiler instance to report errors.

This API is a public API for projects registering custom functions. However, it is also used internally (it is necessary to unlock #266).
For now, the warning message is logged as is. A future improvement once #303 is solved will be to inject the trace automatically for all warnings, as done in dart-sass. I even already implemented it without including it in this PR due to sass-spec failures (where we would no longer match the libsass warnings due to adding a trace but not matching the dart-sass ones due to having a different trace format), which is the reason why I opened #303.
The architecture using a globally available function which forwards to a callback set by the compiler while it runs is inspired from dart-sass, which does something similar.

For now, the Compiler is still useful  in custom function if you want to use all the assertion helpers. However, these helper don't really need to be in the Compiler as they don't depend on the compiler instance (except to access the other helpers). So they could be moved (with a BC layer) into static methods on a `ValueHelper` class (already mentioned in https://github.com/scssphp/scssphp/issues/309).
Once #265 lands, all the assertion methods of the new API will be on the Value classes directly instead (see #301 as they are already implemented).

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -484,9 +484,18 @@ class Compiler
             $this->rootBlock = null;
             $this->rootEnv   = $this->pushEnv($tree);
 
-            $this->injectVariables($this->registeredVars);
-            $this->compileRoot($tree);
-            $this->popEnv();
+            $warnCallback = function ($message, $deprecation) {
+                $this->logger->warn($message, $deprecation);
+            };
+            $previousWarnCallback = Warn::setCallback($warnCallback);
+
+            try {
+                $this->injectVariables($this->registeredVars);
+                $this->compileRoot($tree);
+                $this->popEnv();
+            } finally {
+                Warn::setCallback($previousWarnCallback);
+            }
 
             $sourceMapGenerator = null;
 
@@ -4045,7 +4054,7 @@ class Compiler
             $fname = $this->getPrettyPath($this->sourceNames[$this->sourceIndex]);
             $line  = $this->sourceLine;
 
-            $this->logger->warn("$warning\n         on line $line of $fname", true);
+            Warn::deprecation("$warning\n         on line $line of $fname");
         }
 
         $out = [Type::T_COLOR];
@@ -7378,7 +7387,7 @@ class Compiler
             $name = $this->compileStringContent($this->coerceString($functionReference));
             $warning = "Passing a string to call() is deprecated and will be illegal\n"
                 . "in Sass 4.0. Use call(function-reference($name)) instead.";
-            $this->logger->warn($warning, true);
+            Warn::deprecation($warning);
             $functionReference = $this->libGetFunction([$functionReference]);
         }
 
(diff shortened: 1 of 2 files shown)


stof:
@robocoder @Cerdic what do you think about the naming of this API ?
For now, I used `Reporter::warn` to avoid the repetition of `Warning::warn`.
I also intentionally avoided to expose that as `Compiler::warn`, as that would make it easier to couple things to the actual Compiler parts again by mistake.

stof:
@robocoder @Cerdic can you give me feedback on the naming of this API ? I'd like to merge this to unlock moving forward (I'd like to actually write some proper documentation about extending scssphp through supported extension points)

Cerdic:
Wondering if `Reporter` is not a bit too generic for this specific goal... What about `Alert::warn`? 
Beside that no real better proposal - but I'm not the best name-finder

robocoder:
I agree on having an API but I'm neutral on the name. (I associate "reporter" with "news" IRL, so "warn" is loaded, eg "sensationalize)

stof:
@Cerdic I agree about `Reporter` being too generic. I'm not satisfied with the name. But I needed to use _something_ to request feedback. I think that `Alert::warn` might already been better, even though `alert` seems too strong).

For reference, in dart-sass, the module is named `warn.dart` and exposes a `warn` function. But the usages of the module are importing it without namespace (avoiding a `warn.warn` weirdness).
This is not really an option for us. We could make that a function rather than a static method, but we would still need the static method internally (we don't want to rely on a global variable to access the callback), and it would be painful due to the lack of function autoloading in PHP (and the fact that `use function` is not supported in the older versions we target). That's because PHP does not have the same concept of modules than dart (or EcmaScript or Sass for that matter)

Cerdic:
Maybe `Informer` ?
(I went also through `Scribbler`, `Scribe`, `Discloser` and `Peeper`…)

robocoder:
Messenger?

stof:
I looked at usages of the warn API in dart-sass, and more than 90% of these usages are for deprecations rather than other warnings.
The non-deprecation warnings are found are:
- the implementation of `@warn` (obviously)
- a warning when using a color in interpolation, as that's likely a mistake about not recognizing that the identifier-like value was a named color and not an unquoted string
- a warning about having a selector without any property in the indented syntax (so not relevant for us as we don't parse it) as it won't be rendered at all
- a warning when finding `&&` in the parser (as that's not a `and` but 2 parent selectors)

Everything else are deprecation warnings. In particular, all usages in the SassScript layer are about deprecations, and that means that it is likely to be the same for projects using that API in their custom importers or custom functions.

Base on that, here is a new naming proposal: `Warn::deprecation` and `Warn::warning` (I'm open to ideas if you have alternatives for the second one, but I'm fine with it looking less good, as it will be less used).

What do you think ?

stof:
@robocoder @Cerdic I updated the implementation with my latest naming proposal

PR #336, opened 2021-03-13
Change the visibility of new color helpers to private

stof:
This avoids exposing them as a public API that needs to be covered with backward compatibility.


stof:
these helpers have been introduced when implementing support for HWB, which is not released yet.

PR #356, opened 2021-05-01
Deprecate returning a PHP value from function implementations

stof:
This enforces that functions return a proper sass value instead.
Returning null is still allowed as this is used as a special case to render a css function instead. This case will stay supported until the refactoring of values and function signatures.

This avoids using the guessing-based value coercion for returned values. Functions must manage themselves what they return (custom functions might decide to use the ValueConverter from #355 but core functions always implement the logic directly as they know what they want to return).

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -796,6 +796,11 @@ parameters:
 			path: src/Compiler.php
 
 		-
+			message: "#^Method ScssPhp\\\\ScssPhp\\\\Compiler\\:\\:getTypeOf\\(\\) has parameter \\$value with no value type specified in iterable type array\\.$#"
+			count: 1
+			path: src/Compiler.php
+
+		-
 			message: "#^Method ScssPhp\\\\ScssPhp\\\\Compiler\\:\\:getVariables\\(\\) return type has no value type specified in iterable type array\\.$#"
 			count: 1
 			path: src/Compiler.php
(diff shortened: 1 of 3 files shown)


stof:
I'd like feedback on that deprecation. With the value refactoring coming in 2.x (probably in 2.1), this could mean that functions have to be updated twice, once to add the `ValueConverter::fromPhp` or use the array-based value API, and then during the migration to the new system.
We might want to avoid that step, keeping the value coercion logic in place for the current way of dealing with values.

stof:
Note that even if we don't deprecate that, I would still merge the part changing core functions.

mahagr:
I'm fine with this change, though I've not tested it. We return a string in a few places (mostly fonts and handling url-like php streams), so maybe some documentation about how to use the value types would be nice to have as I've never really figured out how to build scss tree from it.

stof:
@mahagr in your case, the simple way would probably be to use the new `ValueConverter::fromPhp` API.

Writing some docs about about the representation of values would be helpful indeed. But it won't be easy as it is quite messy (if we restrict to document what custom functions can receive, it might be a bit easier). That's one of the main reasons pushing me to work on #265 for 2.1

stof:
@mahagr my main concern about introducing this deprecation is the fact that this would force projects to refactor their custom functions twice: in 1.5 to avoid relying on the automatic conversion and in 2.1 to migrate to the new value API.

mahagr:
Maybe the best approach would be to ask projects to use the new `ValueConverter´ in custom functions (for PHP native returns) and make sure it keeps on working when you convert the code to use the new value types (usually string). For the rest, they can keep on using the current array structure until there will be a replacement. It may also be a good idea to have a conversion for the types when you start using new values internally.

Maybe even introduce `ValueConverter::fromOldApi()`, which takes in the old values and later converts those to use the new API? That would allow them to keep working and thus, gives some time for the developers to update their code...?

stof:
@mahagr when migrating to the new API, we will ask projects to refactor code (with an opt-in at the time of registering the function). Modern functions will receive a list of Value objects as arguments and will be required to return a Value object. Using `ValueConverter::fromPhp` would be possible, but discouraged in that context (in favor of working with the Value API directly, which will be nicer to work with). We will still support legacy functions (with a deprecation warning) at that point.

stof:
@robocoder @Cerdic what do you think about that ? Should we introduce that deprecation for custom functions in 1.5 or should we rather allow to keep relying on the automatic conversion until the migration to the new API in 2.1 ?

robocoder:
Better to deprecate/remove early. Adding functionality later, generally, isn't a compat buster.

PR #364, opened 2021-05-04
Deprecate silent interpolation errors in multiline comments

### LINKED ISSUE #358: Deprecate non-standard support for broken interpolation in comments
Sass loud comments support interpolation. When implementing that to fix #38 in 1.0.5, a no-standard fallback was kept, rendering the comment source as is in case of error during the evaluation.
this non-standard behavior should be deprecated (reporting a Sass warning) so that we can then remove support for it.

stof:
Closes #358

I also submitted https://github.com/sass/sass-spec/pull/1644 to cover that case in sass-spec (which we will need to skip until the 2.0.0 release as sass-spec expects an error, not a deprecation warning while ignoring the error).

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -13,6 +13,8 @@
 namespace ScssPhp\ScssPhp;
 
 use ScssPhp\ScssPhp\Exception\ParserException;
+use ScssPhp\ScssPhp\Logger\LoggerInterface;
+use ScssPhp\ScssPhp\Logger\QuietLogger;
 
 /**
  * Parser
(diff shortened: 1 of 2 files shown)


src/Parser.php
```diff
@@ -1580,7 +1592,14 @@ protected function whitespace()
                 } else {
                     $comment[] = $c;
                     $staticComment = substr($this->buffer, $startCommentCount, $endCommentCount - $startCommentCount);
-                    $this->appendComment([Type::T_COMMENT, $staticComment, [Type::T_STRING, '', $comment]]);
+                    $commentStatement = [Type::T_COMMENT, $staticComment, [Type::T_STRING, '', $comment]];
+
+                    list($line, $column) = $this->getSourcePosition($startCommentCount);
+                    $commentStatement[self::SOURCE_LINE] = $line;
+                    $commentStatement[self::SOURCE_COLUMN] = $column;
+                    $commentStatement[self::SOURCE_INDEX] = $this->sourceIndex;
```
stof, line 1600: This ensure that errors happening in comments can be associated to the right position in the source file. I'm doing that only for cases that contain an interpolation (static comments are in the other branch of the conditions) as static comments won't need it (there is nothing that can trigger an error in them).

PR #372, opened 2021-05-04
Refactor the implementation of sass argument lists

stof:
Argument lists are expected to be a list containing positional arguments, with a way to access named arguments. The representation of the value was mixing the named arguments with the positional ones in the list, which was breaking the usage as a list (not covered by sass-spec) and also returning wrong data in the `keywords` function.
This is fixing bugs for the usage of an argument list in Sass code.
This is technically a BC break for custom host function that expect to accept arbitrary named arguments while declaring arguments explicitly. However, this is considered OK. This representation of argument lists is not documented anywhere. And in previous releases of scssphp, there was no such usages in the core either to allow discovering how to do it (change-color, scale-color and adjust-color) were not implemented in terms of a rest argument). The documentation about implementing custom functions was showing how to achieve this for a function without an argument declaration only. So it is unlikely that anyone discovered how to do it in a way impacted by the BC break.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -766,6 +766,11 @@ parameters:
 			path: src/Compiler.php
 
 		-
+			message: "#^Method ScssPhp\\\\ScssPhp\\\\Compiler\\:\\:getArgumentListKeywords\\(\\) has parameter \\$value with no value type specified in iterable type array\\.$#"
+			count: 1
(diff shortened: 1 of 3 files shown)


stof:
@Cerdic @robocoder are you OK with this small BC break for the minor version ?

robocoder:
Lgtm

PR #386, opened 2021-05-05
Upgrade sass-spec

stof:


tests/specs/sass-spec-exclude.txt
```diff
@@ -258,6 +258,7 @@ core_functions/map/values/empty
 core_functions/map/values/single
 core_functions/math/round/up/within_precision
 core_functions/meta/call/error/invalid_args
+core_functions/meta/content_exists/controls/false
```
stof, line 261: `content_exists` is not implemented at all in scssphp currently. So we of course fail the new test as well.

PR #394, opened 2021-05-13
Use PHP 8.0 for the windows and macOS CI jobs

stof:
The github actions environment comes with PHP 8.0 preinstalled on windows and macOS (on Linux, it has both 7.4 and 8.0), which makes the setup faster than having to install another version.


stof:
In https://github.com/scssphp/scssphp/pull/391/checks?check_run_id=2575236910, the `Setup PHP` step took 1min 17s on macOS. Here, thanks to reusing the preinstalled version, it tool 2s.

PR #396, opened 2021-05-14
Try running tests on PHP 8.1

stof:
I don't plan to merge this yet, as PHP 8.1 is not ready yet. But this would give us a first indication of our compatibility.


stof:
closing this as I won't enable it in CI until PHPUnit supports it officially (which won't happen until PHP 9.1 stabilizes)

PR #406, opened 2021-05-17
Fix bootstrap compilation

### LINKED ISSUE #403: v1.5.1 fails to compile Bootstrap 4.6.0
I looks like my issue with Bootstrap 4.6 is unrelated to the Bootstrap 5 issue (https://github.com/scssphp/scssphp/issues/397#issuecomment-841632120).

I updated to v1.5.1 and I'm still getting this error:
```
/* SCSS compile error: $n: List index may not be 0.: extensions/Bootstrap/resources/bootstrap/scss/mixins/_box-shadow.scss on line 13, at column 9
Call Stack:
#0 mixin box-shadow extensions/Bootstrap/resources/bootstrap/scss/_custom-forms.scss on line 36
#1 import extensions/Bootstrap/resources/bootstrap/scss/_custom-forms.scss (unknown file) on line 1*/
```

stof:
Closes #403 

This reverts #380. fixing spec compliance for lists to avoid breaking list functions is too much work, as it looks like it requires fixing interpolation first.
As Bootstrap seems to compile fine without the `nth` refactoring (even though some other list-related code might have bugs that we already had in 1.4 due to not being spec compliant), I prefer delaying the spec compliance fix to a 2.x release rewriting the parser.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,6 +41,7 @@
         "squizlabs/php_codesniffer": "~3.5",
         "symfony/phpunit-bridge": "^5.1",
         "twbs/bootstrap": "~5.0",
+        "twbs/bootstrap4": "4.6.0",
         "zurb/foundation": "~6.5"
     },
     "repositories": [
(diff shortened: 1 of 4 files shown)


GrupoWebex:
It is working now, thanks for your support!

stof:
> It is working now, thanks for your support!

I know it is: I added it in the CI as part of the fix

PR #407, opened 2021-05-18
Prepare the 1.5.2 release

stof:


mahagr:
Looks like this version works without any issues in Gantry 5.5, thank you for all your work!

PR #408, opened 2021-05-18
Drop support for PHP older than 7.2

### LINKED ISSUE #314: [RFC] Min supported version for 2.0
This is restarting the discussion of #202 based on new changes in downstream needs that blocked the decisions previously.

Note that this discussion is about the **minimum** supported version of PHP. The range of supported version will go from that minimum version up to the current PHP version. so downstream projects that already requires a higher minimum version than us are totally fine. The issue is if they want to keep support for a version *lower* than our minimum.
Also note that this change will apply only to 2.0. Our maintenance branch for 1.x will still support PHP 5.6+, but it won't receive anything else than critical bug fixes once 2.0 is out. That's still a solution for maintenance branches of downstream projects though if needed.

## Interesting features

The minimum supported PHP version will affect our ability to use these features in our own code:

- scalar types: PHP 7.0
- return types: PHP 7.0
- null-coalescing operator: PHP 7.0
- nullable types: PHP 7.1
- variance for signatures: PHP 7.2 (not needed for the Value objects, but useful if we want to port the dart-sass AST visitors later to allow using native types)
- typed properties: PHP 7.4 (nice-to-have, as our properties will stay private anyway to enforce immutability)
- union types: PHP 8 (useful for numbers, as they are `int|float`, i.e. the dart `num` type)
- enums: PHP 8.1 (could be nice to represent some parts of the AST or list separators for instance, but far too new to even consider PHP 8.1)

PHPUnit 8 is the first PHPUnit version which includes Life support, i.e. the fact that it will be made compatible with new PHP versions even after its end of maintenance. It requires PHP 7.2+. Being able to use PHPUnit 8+ would then make it much easier to maintain our tests (as we might not need to make tests compatible with a wide range of PHPUnit major versions to satisfy our PHP needs).

PHP 7.3 does not unlock any new feature by itself. So there is no reason to make it the min version (if we need to keep support for 7.3, we can keep support for 7.2 as well for the same maintenance cost).

## Downstream requirements

Top dependencies based on https://packagist.org/packages/scssphp/scssphp/dependents?order_by=downloads

- October CMS: 7.2.9
- Contao: 7.3
- Shopware: 7.2 (7.4 in their dev version)
- [Typo3 Bootstrap package](https://packagist.org/packages/bk2k/bootstrap-package): 7.2
- EZPlatform: 7.3
- OpenCart: 7.3 (taking the requirement from the master branch, as the latest tag is from 2018 instead and still uses `leafo/scssphp`)
- [Sass compiler for Typo3](https://packagist.org/packages/svewap/ws-scss): 7.2
- mediawiki/scss: 5.6  in the composer package. But Mediawiki itself now requires PHP 7.3

Extra known downstream projects:
- [SPIP](https://git.spip.net/spip-contrib-extensions/scssphp): 7.1 or 7.3 (see https://github.com/scssphp/scssphp/issues/310#issuecomment-796615476)
- Gantry 5: 5.6
- Grav (supported at least through Gantry, maybe by other themes too): 7.3
- Joomla 3 (supported at least through Gantry, maybe by other themes too): 5.3 (recommends 7.3+)
- Joomla 4: 7.2.5 (recommends 7.4+)
- Wordpress: 5.6. Recommends 7.4+

stof:
Closes #314

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [ '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0' ]
+                php: [ '7.2', '7.3', '7.4', '8.0' ]
                 extensions: [ '' ]
                 os: [ubuntu-latest]
                 include:
--- a/bin/pscss
+++ b/bin/pscss
@@ -13,8 +13,8 @@
 
 error_reporting(E_ALL);
 
-if (version_compare(PHP_VERSION, '5.6') < 0) {
-    die('Requires PHP 5.6 or above');
+if (version_compare(PHP_VERSION, '7.2') < 0) {
+    die('Requires PHP 7.2 or above');
 }
 
 include __DIR__ . '/../scss.inc.php';
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "psr-4": { "ScssPhp\\ScssPhp\\Tests\\": "tests/" }
     },
     "require": {
-        "php": ">=5.6.0",
+        "php": ">=7.2",
         "ext-json": "*",
         "ext-ctype": "*"
     },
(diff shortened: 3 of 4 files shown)


stof:
this does not modernize the codebase yet to add types. I'll do it in a separate PR after removing deprecated code.

PR #424, opened 2021-05-29
Implement the stylesheet parser ported from dart-sass

### LINKED ISSUE #413: The parser is not resilient against invalid input
As an experiment, I played with https://github.com/nikic/PHP-Fuzzer on this library.
I did not went far into the experiment yet, but this already uncovered several cases where the parser tries to access some offsets in the source without checking whether they are past the end of the input or no. for instance, `p:` as input triggers such notice instead of triggering a SassException about the input being invalid (if you don't fail on PHP notices, it will still report the parse error after that).

For reference, here is the fuzz target I used:

```php
<?php

use ScssPhp\ScssPhp\Compiler;
use ScssPhp\ScssPhp\Exception\SassException;
use ScssPhp\ScssPhp\Logger\QuietLogger;

require __DIR__ . '/../../scss.inc.php';

$fuzzer->setTarget(function(string $input) {
    $compiler = new Compiler();
    $compiler->setLogger(new QuietLogger());
    $compiler->setImportPaths([]);

    try {
        $compiler->compileString($input);
    } catch (SassException $e) {
        throw $e;
    } catch (Exception $e) {
        throw new Error('An exception was thrown', 0, $e);
    }
});
```

stof:
This is a full rewrite of the stylesheet parser by porting the dart-sass one (`@use` and `@forward` are not implemented for now, but usage of namespaced members is parsed fully before throwing a proper error about modules not being implemented yet).

This new parser outputs a proper Sass AST. I also ported part of the SourceSpan Dart package to represent spans of the original source. This will allow us to have a much better error reporting in the future (but I haven't implemented the rendering of code snippets associated to the spans yet, this is left for a next PR).

TODOs:
- [ ] use this new parser in the Compiler
- [ ] figure out the impact on the parser cache
- [ ] figure out how to deal with interpolation (converting the Sass AST to our legacy representation to delay rewriting the evaluation layer might not be possible, as the legacy way to represent interpolation is totally flawed)
- [ ] remove the old parsing code
- [x] deal with phpstan reporting a few weird things

Closes #413

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,6 +1,16 @@
 parameters:
 	ignoreErrors:
 		-
+			message: "#^PHPDoc tag @template T for class ScssPhp\\\\ScssPhp\\\\Ast\\\\Sass\\\\Statement\\\\ParentStatement with bound type array\\<ScssPhp\\\\ScssPhp\\\\Ast\\\\Sass\\\\Statement\\> is not supported\\.$#"
+			count: 1
+			path: src/Ast/Sass/Statement/ParentStatement.php
+
+		-
+			message: "#^PHPDoc tag @template T for class ScssPhp\\\\ScssPhp\\\\Ast\\\\Sass\\\\Statement\\\\ParentStatement with bound type null is not supported\\.$#"
+			count: 1
+			path: src/Ast/Sass/Statement/ParentStatement.php
+
+		-
 			message: "#^Property ScssPhp\\\\ScssPhp\\\\Block\\:\\:\\$children type has no value type specified in iterable type array\\.$#"
 			count: 1
 			path: src/Block.php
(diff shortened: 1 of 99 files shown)


stof:
While looking at doing the next steps of that work (actually using the new parser), I figured out that it cannot be yet easily. The StylesheetParser does not parse media queries or selectors at all (as they allow interpolation *before* parsing). Migrating to the new parser involves extra steps:
- implement the MediaQueryParser (easy. Its only a few LOC and I have a branch locally with this code)
- implement the SelectorParser and the selector AST
- once those are available, we can rewrite our compiler to use them, which will require rewriting some parts of the logic (basically, most places where the AST contains interpolation)

to keep PRs at a manageable size, I suggest splitting the work into multiple PRs instead of doing everything here.

Issue #256, opened 2020-11-07
Investigate native support for stream wrappers for import paths


stof:
#214 reports than Gantry 5 overrides the importer-related methods to use their `gantry-theme` stream wrapper to resolve file.
We should look at a way to support such use case through official extension point, i.e. `addImportPath` with either a string or a callable. If that cannot be done directly, we might need to introduce a more powerful `Importer` API to support that.

mahagr:
@stof I've been reworking the code to use callables to get the real file paths and it seems to work so far (except for map files having absolute paths for source files, which I'm still looking at).

I don't think that SCSS compiler needs to add support for streams, they work out of the box even without the callable (I just tested it).

stof:
@mahagr is this using the unmodified compiler or your customized one which changes the handling of imports to use `$this->streamNames` ?

mahagr:
There is no need to have the variable, it was just used to fix some of the issues I ran into with the old code. Gantry 5 code was originally written 5 years ago for the very old compiler. I had tried to keep it somewhat updated, but with this rewrite I am not even passing streams to SCSS compiler as they are resolved in the callable like you suggested.

That said, even the callable is not really needed to handle streams -- `realpath()` works just fine with local PHP streams. 

The only thing I cannot do anymore is to use `$compiler->getParsedFiles()` to determine if recompilation is needed (it has several issues), but I don't think it matters too much as I can likely deal with that inside my callable.

stof:
> The only thing I cannot do anymore is to use `$compiler->getParsedFiles()` to determine if recompilation is needed (it has several issues)

Can you describe these issues ?

> That said, even the callable is not really needed to handle streams -- `realpath()` works just fine with local PHP streams.

Grat news. This means this could be considered resolved...

mahagr:
We already discussed the issue with `getParsedFiles()` remembering the previous compilations and the issue it causes with caching. The second issue is also caching related as the timestamp checks as the paths are real paths and do not take into account overrides, which was the most requested feature for Gantry -- automatically recompile when user overrides the source files in a different location.

I consider this issue closed as well. We can go through the caching issues in a separate issue.

stof:
@mahagr you mean that the stream can resolve to different files ? would calling `filemtime` on the stream_wrapper path work for overrides ? Maybe that would be the need to natively support stream wrappers properly.

Cerdic:
I opened a new issue for the cache invalidation confitions #268

mahagr:
[...]

PR #448, opened 2021-08-04
Enhancing custom import functions

ytterx:
Hi, I was encountering an issue where I wanted to import some parts of a thirdparty package (bootstrap) but that was checkout in such a way that it not really nice to read in my scss source files.
A bit context, i'm making a template engine which uses scss, but allows you to import parts of bootstrap.

What I wanted to work:
`@import "bootstrap/functions";`
The path of this functions file is:
`bootstrap\scss\_functions.scss`

With these changes, the `functions` part is now resolving to `_functions.scss`

I needed access to the `Compiler->findImport` function for this, so I'm not sure how this will work with the import caching mechanism. Or how this could be improved.

As a bonus, I force the input and output to have Linux style line-endings as I was having problems with the unittests on Windows.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,3 +11,6 @@ phpstan*.neon export-ignore
 phpunit.xml.dist export-ignore
 tests/ export-ignore
 vendor-bin/ export-ignore
+tests/outputs/*.css text eol=lf
+tests/inputs/*.scss text eol=lf
+tests/inputs/imports/*.scss text eol=lf
--- a/docs/docs/extending/importers.md
+++ b/docs/docs/extending/importers.md
@@ -36,3 +36,26 @@ $compiler->addImportPath(function($path) {
 // will import 'stylesheets/sub.scss'
 echo $compiler->compileString('@import "sub.scss";')->getCss();
 ```
+
+Optional, the current `\ScssPhp\ScssPhp\Compiler` can be accessed with the
+second parameter. Allowing to retry the `findImport` function after correcting
+`$path`
+```php
+use ScssPhp\ScssPhp\Compiler;
+
+$compiler = new Compiler();
+$compiler->addImportPath(function($path, $compiler) {
+    if (Compiler::isCssImport($path)) {
+        return null;
+    }
+
+    if (substr($path, 0, 19) !== "thirdparty_package/") {
+        return null;
+    }
+
+    return $compiler->findImport(substr($path, 19), "path_to_thirdparty_package/");
+});
+
+// will import 'path_to_thirdparty_package/_variables.scss'
+echo $compiler->compileString('@import "thirdparty_package/variables";')->getCss();
+```
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -5294,7 +5294,7 @@ EOL;
      *
      * @return string|null
      */
-    private function findImport(string $url, ?string $currentDir = null): ?string
+    public function findImport(string $url, ?string $currentDir = null): ?string
     {
         // Vanilla css and external requests. These are not meant to be Sass imports.
         if (self::isCssImport($url)) {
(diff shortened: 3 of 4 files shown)


stof:
I'm -1 for passing the compiler as second argument to the callback, as this would forbid us from moving this logic outside the Compiler class in the future (which is definitely planned as part of the v2 work that moves the architecture closer to the dart-sass one to ease the maintenance while being compatible.

I'm not sure either about exposing `Compiler::findImport` as a public API. The equivalent API is not exposed in dart-sass and there is no such shortcut for implementing the loading behavior you describe here in either the dart or JS APIs of dart-sass.

ytterx:
> @ytterx @slprime Assuming findImport() remains private, how is the existing callback functionality not meeting your needs?

The thing is that for example I have bootstrap located somewhere which I would like to use, bootstrap uses files like _functions.scss

But a potential conflict might occur if I have my own functions.scss or _functions.scss. So for my functionality, I purposely make a sort of namespace "bootstrap/" to allow me to dectect with the call back if I can resolve the path for bootstrap. Just like done in the examples.

So keeping findImport private, (more over not accessable) I need to implement findImport myself, maybe causing me to run out of spec if any update might happen.

Allowing the functionality of findImport in a sort of "final" library function / class allows developers and me to re-use the algorithm as implemented and keeping within spec.

stof:
@ytterx resolving `@import "bootstrap/functions"` to `bootstrap/dist/scss/_functions.scss` is *already* going out of spec.

ytterx:
> @ytterx resolving `@import "bootstrap/functions"` to `bootstrap/dist/scss/_functions.scss` is _already_ going out of spec.

But does this library needs to be so strict in reducing it's usability to use the functionality it already provides? It seems ridiculous to me that I would have to implement exactly the same code for something which in my opinion should be more generic.

Again I see why you might not want to provide the Compiler in the call back itself (although that was done in the past?).

The callback allows me to kind of extend the spec to my needs, but why should I be limited to not re-use this part of the spec?

stof:
@ytterx the issue is that if we expose this logic to be reusable, we cannot refactor it anymore without breaking BC.
As we are in the process of doing a full rewrite of the internals of the library (to actually become spec compliant), anything that forbids refactoring has a *huge* cost in term of maintenance.

ytterx:
> @ytterx the issue is that if we expose this logic to be reusable, we cannot refactor it anymore without breaking BC.
> As we are in the process of doing a full rewrite of the internals of the library (to actually become spec compliant), anything that forbids refactoring has a _huge_ cost in term of maintenance.

Then please consider this in the new "design" with the refactoring. I don't see a reason why BC needs to be broken if the code is properly designed. Basicly allow from a callback to either continue searching in a new path (with the partial file search functionality) or to call a function which returns the (possible) file path(s)

stof:
> I don't see a reason why BC needs to be broken if the code is properly designed.

that's the whole point. The `Compiler::findImport()` method is not what I consider "properly designed". In the new architecture, the compiler itself will do almost nothing, being only responsible for coordinating the various other classes. And so that logic will move elsewhere. If we need to maintain BC on `Compiler::findImport()`, it prevents us from moving this logic elsewhere (and then potentially also prevents moving the caller logic, etc...)

mahagr:
BTW, it would be really cool if the future classes with the logic could be extended/replaced as long as they follow the interface. So please give a way to pass the classes during initialization and avoid using the final keyword in methods that implement parts that could be overridden. I don't care if I needed to do more work to keep my logic up to date with the library as long as there's a warning if the methods may change and that you mark which methods are part of external API and which are not.

In my case, some of the SCSS files may be located elsewhere, even in a database.

stof:
> In my case, some of the SCSS files may be located elsewhere, even in a database.
> 
> Of course, I could also add a stream wrapper (in fact, I already have it), but even in that case the code needs to be compatible with those, meaning no `realpath()` calls are allowed etc..

Supporting files coming from elsewhere than the filesystem is indeed something I have in mind (see #256 tracking it by talking about stream wrappers, even though that might not be the final solution).

> it would be really cool if the future classes with the logic could be extended/replaced as long as they follow the interface. So please give a way to pass the classes during initialization and avoid using the final keyword in methods that implement parts that could be overridden.

the architecture I have in mind (partially) would indeed allow extension points. But those would probably not be based on inheritance.

PR #450, opened 2021-08-11
add import once functionality again

crayssnlabs:
Hi!

I have re-added the 'import once' functionality to optimize the on the fly compiling. For IDEs (like phpStorm) the autocompletion will be improved if there is an util/mixing import in the current scss file. This necessary imports are now ignored and will not slow down the compiling process.

Your can enable/disable the functionality. (disabled by default) 

I have also refactored the list() to [].

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -110,10 +110,23 @@ final class Compiler
      * @var array<int, string|callable>
      */
     private $importPaths = [];
+
     /**
      * @var array<string, Block>
      */
     private $importCache = [];
+
+    /**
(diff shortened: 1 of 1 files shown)


stof:
This implementation is even worse than the one we deprecated in 1.5 and removed from the 2.0-dev codebase. Instead of implementing it as a new non-standard directive, it changes the behavior of the standard `@import` to be non-compliant with the spec. This is a no-go.

crayssnlabs:
Thank you for your answer. We will use the non-standard fork with improved functionality. Thank you for your work!

PR #458, opened 2021-09-06
Improve the handling of charset rules

stof:
- Fix the handling of `@charset` in source files, ignoring it properly
- Implement the `charset` option that was added in dart-sass recently to allow disabling the rendering of the `@charset "UTF-8";` prefix
- ~~Use an UTF-8 BOM as the prefix in compressed mode as it is shorter (this is what dart-sass does).~~ (not applied in 1.x)

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -221,6 +221,8 @@ class Compiler
     protected $storeEnv;
     /**
      * @var bool|null
+     *
+     * @deprecated
      */
     protected $charsetSeen;
     /**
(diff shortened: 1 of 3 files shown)


mahagr:
Wow, thank you! I just ran into this issue and had to remove those from the files.

stof:
hmm, maybe I can submit that to 1.x rather than master to release it sooner. The 2.0 release will take a bit longer as I discovered that our bad implementation of interpolations in 1.x makes it a lot harder to do an incremental rewrite.

stof:
The commit about using the UTF-8 BOM won't work for 1.x, as 1.x does not rely on an OutputStyle for its output (we have a setter for it, but internally we still need to handle all the deprecated formatters)

mahagr:
If it's possible, I would love to get this implemented in 1.x. If not, I've already worked around it, so it can also wait.

stof:
this is now targetting the 1.7.0 release, as I need to do a 1.x release anyway to fix a bug.

The usage of the UTF-8 BOM will be implemented separately for the master branch.

PR #464, opened 2021-09-15
Improve warning assertions in the sass-spec tests

stof:
- support warnings from alternative implementations (our warnings often match the libsass ones)
- assert the absence of warnings for tests without an expected warning

This is especially relevant due to https://github.com/sass/sass-spec/pull/1694 changing the default warning message to be the dart-sass one.


tests/specs/sass-spec-exclude-warning.txt
```diff
@@ -81,6 +71,7 @@ libsass-closed-issues/issue_308
 libsass-closed-issues/issue_674
 libsass-closed-issues/issue_759
 libsass-todo-issues/issue_1096
+libsass-todo-issues/issue_1798/3
```
stof, line 74: Those 2 added tests are tests that trigger the deprecation warning `DEPRECATION WARNING: Unterminated interpolations in multiline comments are deprecated and will be removed in ScssPhp 2.0`.
Those are false positives of the warning, because those comments are not preserved in the scss output (and so interpolation will not be parsed in them)

stof: and now fixed, as I found the issue

PR #497, opened 2021-09-21
Abstraction for Filesystem

lausek:
This pull request tries to encapsulate filesystem access as described in https://github.com/scssphp/scssphp/issues/256.

**NOTE:** Importing does not work as `$compiler->currentDirectory` is null. `$compiler->addImportPath("")` can be used as a workaround. How should this be handled?

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -19,6 +19,8 @@ use ScssPhp\ScssPhp\Exception\CompilerException;
 use ScssPhp\ScssPhp\Exception\ParserException;
 use ScssPhp\ScssPhp\Exception\SassException;
 use ScssPhp\ScssPhp\Exception\SassScriptException;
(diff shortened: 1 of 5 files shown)


stof:
Our goal for one of the 2.x releases (not sure it will be in 2.0) is to improve the importer abstraction. So I'm -1 about this FileReader abstraction which we would then have to maintain forever (or at least until 3.0)

lausek:
> Our goal for one of the 2.x releases (not sure it will be in 2.0) is to improve the importer abstraction.

Unfortunately, I am not familiar with the plans on improving the importer abstraction. At the moment, I need a solution to import CSS files without the need to hit the filesystem. The `FileReader` approach is a simple but also flexible solution for this task. I will stick to my fork until the importer feature has been stabilized.

mahagr:
I'm also using my own version right now, for the same reasons.

PR #511, opened 2021-11-11
Implement the CSS AST and its serialization

stof:
This implements the CSS AST ported from dart-sass as well as the associated serialization (without sourcemap support for now, but changes will be small to add it)


stof:
I initially tried integrating the new parser into the existing compiler by converting the Sass AST into our legacy structure. But that was too painful, given that our legacy parsing produces a totally broken output in some cases. So I decided to port the CSS AST instead, to perform the rewrite of the evaluation step as a next step (instead of spreading it over 2.x minor versions).

PR #517, opened 2021-12-14
Fix php8.1 strpos deprecated message for pure css imports

basteyy:
In case you import a css file `$filename` is null. That causes the following warning:

```
Deprecated: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in /var/www/html/vendor/scssphp/scssphp/src/SourceMap/SourceMapGenerator.php on line 355
```

By adding a check if `$filename` is set (and not null), the provided null-value to strpos is avoided.

### DIFF
--- a/src/SourceMap/SourceMapGenerator.php
+++ b/src/SourceMap/SourceMapGenerator.php
@@ -312,7 +312,7 @@ final class SourceMapGenerator
         }
 
         // Remove extra leading path separators.
-        if (strpos($filename, '\\') === 0 || strpos($filename, '/') === 0) {
+        if ($filename && (strpos($filename, '\\') === 0 || strpos($filename, '/') === 0)) {
             $filename = substr($filename, 1);
         }
 


stof:
Can you share a reproducer ? I don't see how it can be `null` here (see the argument type)

basteyy:
> Can you share a reproducer ? I don't see how it can be `null` here (see the argument type)

Sure: https://gist.github.com/basteyy/3762144f0301850f0c0cc64a62b1d65f

stof:
@basteyy can you reformat the error in your gist to preserve the line breaks in the stack trace ? It would make it a lot easier to read.

stof:
@basteyy workaround for you: pass the file path as the second argument of `compileString`. This bug happens only when the path is not known.
Side note: passing the path also makes relative imports work, and it will improve sourcemaps (by marking styles as coming from style.scss rather than stdin), so it is a good idea to pass that argument when your source comes from a file.

stof:
and in any case, this PR is not the right fix (btw, on the master branch, the error happens before reaching that line due to the argument type)

stof:
Closing in favor of #523 which fixes the root cause instead of trying to avoid a symptom (in a bad way as the master branch fails earlier due to native types being added on arguments)

PR #525, opened 2021-12-27
Add explicit return type to avoid Symfony debug warning

shyim:
07:40:37 INFO      [php] User Deprecated: Method "ArrayAccess::offsetSet()" might add "void" as a native return type declaration in the future. Do the same in implementation "ScssPhp\ScssPhp\Node\Number" now to avoid errors or add an explicit @return annotation to suppress this message.

### DIFF
--- a/src/Node/Number.php
+++ b/src/Node/Number.php
@@ -147,7 +147,7 @@ class Number extends Node implements \ArrayAccess
     }
 
     /**
-     * {@inheritdoc}
+     * @return bool
      */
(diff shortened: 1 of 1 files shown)


stof:
This should target the 1.x branch.

wouterj:
In fact, given the class is final, these can be made real PHP types in the master branch (when upmerging these changes from 1.x)

stof:
@wouterj no they cannot. The master branch supports PHP 7.2+, and `mixed` is not supported there.

PR #555, opened 2022-02-28
Rework the late initialization of the MixinRule::$content property

stof:
PHP static analyzers don't support properly the pattern of late initialization right now.
The new code defines the property as nullable and not read-only, to allow changing the property later in case it is null. This is less explicit about the intent of the property, but it is supported by the tooling.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,11 +1,6 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Property ScssPhp\\\\ScssPhp\\\\Ast\\\\Sass\\\\Statement\\\\MixinRule\\:\\:\\$content \\(bool\\) in isset\\(\\) is not nullable\\.$#"
-			count: 1
-			path: src/Ast/Sass/Statement/MixinRule.php
-
-		-
 			message: "#^Property ScssPhp\\\\ScssPhp\\\\Block\\:\\:\\$children type has no value type specified in iterable type array\\.$#"
 			count: 1
 			path: src/Block.php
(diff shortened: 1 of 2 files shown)


stof:
Looks like phpstan won't add the concept of late initialized properties, and especially not for properties that don't have a native type. So I'll merge this.

mahagr:
Also, it won't work in PHP 8.1 if you change the type-hint into a PHP 8.1 type. In PHP 8.1 you need to set the type to `?bool` and initialize it in either in the definition or in the constructor. This is why I think that the phpstan approach is the correct way to deal with this issue as it prevents you from breaking the code when you convert it to use the proper type declaration.

stof:
@mahagr that's not true. You need to initialize the property before reading it. But that can happen later than in the constructor (the previous code was 100% valid with a typed property).
Requiring to initialize typed properties in the constructor is a requirement added by psalm (and by phpstan when turning on an opt-in parameter) to let them avoid tracking the uninitialized state of properties. But that's not a rule that actually exist in PHP. And that's precisely what I tried requesting to phpstan (even suggesting to have a `@late` phpdoc to opt-in for the full tracking instead of forcing it for all properties).

mahagr:
Hmm, could be that I got the idea from phpstan in that case. Thanks for correcting me as I do find the feature useful.

stof:
checking for uninitialized typed properties must use `isset(...)`, not `null !== ...`

mahagr:
Yup. I think (on my quick tests) you can also `unset()` the variables instead of setting them to be null.

PR #556, opened 2022-03-02
Properly clean the output buffer when an error occurs during formatting

stof:
This makes the debugging of errors a lot easier, as it avoid leaking the content of the buffer to the output.


Hackwar:
Even with this change, I get an assert error on line 360 (formerly line 352)

PR #585, opened 2022-07-02
Allow to replace tilde sign with given path value to fix the tilde sign

ghanu119:
Allow to replace tilde sign with given path value to fix the tilde sign path resolution

Now Tilde( ~ ) can be replaced with the varaibles `path_for_tilde`.

When there is `~` sign exists in import line then define the replacement like:
`$compiler->addVariables(['path_for_tilde' => 'PATH_TO_REPLACE'])`

**AS PER NEW CHANGES**
With working JS frameworks sometimes some library path starts with the `~` [Tilde] sign and this variable will allow you to set the starting `~` to replace the directory path.
```
$compiler->addVariables([
    'path_for_tilde' => [ 'PATH_TO_REPLACE' ]
]);

```


stof:
Replacing that in all strings does not make any sense and would break spec compliance.

We already have a hook for import paths, where you can pass a callable that performs a custom resolution.

ghanu119:
ImportPath hook is not working when there is a import with `~` sign. As sass-loader mange to replace the ~ sign with the node_modules where this package can't work with ~ import sign and also not looked into the the path that hooked using the importPaths.
For exmple,
`
@import '~bootstrap/scss/bootstrap.scss';
@import '~bootstrap-select/sass/bootstrap-select.scss';
` 
Now importPath hook not work with this type of imports.

stof:
What do you mean by "does not work" ?

Your proposal here impacts **all strings**, not only imports, so it is clearly not something we will merge.

ghanu119:
@stof 
New changes made and fixed as per your suggestion.

**test.scss**

```
@import '~bootstrap/scss/bootstrap.scss';
@import '~test/custom~/demo.scss'; // It will only set the path for the first tilde and not effect on second tilde.

.my-custom {
    .sub-custom {
        font-size: 12px;
        font-weight: bold;
    }
}

```
**PHP CODE**
```
<?php

require_once 'scss.inc.php';

use ScssPhp\ScssPhp\Compiler;
use ScssPhp\ScssPhp\OutputStyle;

$compiler = new Compiler();
$nodeModule = $_SERVER['DOCUMENT_ROOT']  . '\scsstest\node_modules\\';
$compiler->setOutputStyle( OutputStyle::COMPRESSED );
$compiler->addImportPath( $nodeModule );
$compiler->addVariables([
    'path_for_tilde' => [$nodeModule]
]);

echo $compiler->compileString(file_get_contents($_SERVER['DOCUMENT_ROOT']  . '..\scsstest\test.scss' ))->getCss();

```

ghanu119:
@stof 
You have to also merge this PR as it will improve the functionality and able to read files from the node_modules folder and there are lots of projects which use the plugin and there css used inside the project.
Please check this.

stof:
I'm closing this PR as I haven't changed my mind since my previous comment.

PR #616, opened 2022-09-03
Add an API for location-aware loggers

stof:
This new interface will give more flexibility to loggers to report locations in a structured way.
Once the rewrite is complete, all internal usages of the logger will work with that interface, letting the logger deal with adding the
location in the message for textual output.
The `AdaptingLogger` provides the compatibility layer to support loggers implemented against the 1.x interface.

### DIFF
--- a/src/Logger/AdaptingLogger.php
+++ b/src/Logger/AdaptingLogger.php
@@ -1,0 +1,76 @@
+<?php
+
+/**
+ * SCSSPHP
+ *
+ * @copyright 2012-2020 Leaf Corcoran
+ *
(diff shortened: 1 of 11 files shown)


stof:
@Cerdic one thing I'm not 100% satisfied about is the `LocationAwareLoggerInterface`, but `LoggerInterface` is already taken by the old interface. If you have any idea for a better name, feel free to suggest them

stof:
Thinking about that again, this name actually reflects what is specific about it. So I'll merge it as is (we can still rename it before 2.0 becomes stable if someone suggests a better name)

PR #653, opened 2023-08-24
fix collapse selector on single node

ssltg:
collapseSelectors throw Exception if `@extend` is used with a selector that also is combined with `:is()`

eg:
```scss
.item-header {
    @extend .specific-item-header;
}

:is(.some-table-header, .specific-item-header ) {
    padding: 14px 14px 0;
} 
```

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -1954,6 +1954,11 @@ final class Compiler
             foreach ($selector as $node) {
                 $compound = '';
 
+                if (!is_array($node)) {
+                    $output[] = $node;
+                    continue;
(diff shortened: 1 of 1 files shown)


stof:
Your failing case should be added as a reproducing test in the testsuite (see `tests/inputs` and `tests/outputs`)

stof:
Btw, if this PR should probably target 1.x

ssltg:
> Btw, if this PR should probably target 1.x

made a new one from 1.x https://github.com/scssphp/scssphp/pull/654

PR #654, opened 2023-08-24
fix collapse selector on single node

ssltg:
moved from: https://github.com/scssphp/scssphp/pull/653 to 1.x

collapseSelectors throw Exception if `@extend` is used with a selector that also is combined with `:is()`

eg:
```scss
.item-header {
    @extend .specific-item-header;
}

:is(.some-table-header, .specific-item-header ) {
    padding: 14px 14px 0;
} 
```

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -2090,6 +2090,11 @@ class Compiler
             foreach ($selector as $node) {
                 $compound = '';
 
+                if (!is_array($node)) {
+                    $output[] = $node;
+                    continue;
+                }
+
                 array_walk_recursive(
                     $node,
                     function ($value, $key) use (&$compound) {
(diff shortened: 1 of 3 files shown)


stof:
Can you please make that a new test instead of adding more things in an existing test ?
Having tests with a dedicated focus makes it a lot easier to maintain as we can know what they test. It also makes it easier to compare the test expectations to the official implementation as each test is either matching or no, so a single big test would report a match only if _everything_ is matching the official implementation, instead of having a more precise view with several small tests.
See https://github.com/scssphp/scssphp/pull/574 for instance where I added the new case in a dedicated test instead of adding more stuff in the existing `at_root` test. It requires figuring out a name for the new test, but having an explicit name is precisely what helps understanding what the test is about.

stof (approved):

stof:
Thanks for this fix.
And congrats for the courage of digging into this messy code implementing `@extend` in the 1.x codebase...

PR #657, opened 2023-09-09
Faster Parser->literal()

staabm:
literal is the slowest call when compiling redaxo/redaxo

grafik

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -2626,7 +2626,7 @@ final class Parser
 
         // unicode range with wildcards
         if (
-            $this->literal('U+', 2) &&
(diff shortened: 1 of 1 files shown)


staabm:
error is unrelated, as it also happens in https://github.com/scssphp/scssphp/pull/658 which does not at all change the libs code

stof:
This should be submitted to 1.x. The master branch is in the middle of a whole rewrite and this parser will become unused.

stof:
Btw, this CI error is already fixed in the 1.x branch.

staabm:
thanks for the hint

PR #658, opened 2023-09-09
Use latest phpstan, drop composer-bin-plugin

staabm:


composer.json
```diff
@@ -37,7 +37,7 @@
         "ext-iconv": "Can be used as fallback when ext-mbstring is not available"
     },
     "require-dev": {
-        "bamarni/composer-bin-plugin": "^1.4",
```
staabm, line 40: newer phpstan versions no longer define external dependencies. we don't need the composer bin plugin anymore

staabm:
failling test looks unrelated

composer.json
stof: The main issue is that the 1.x branch supports PHP versions that are not supported by phpstan.

staabm: ohh wow.. PHP5 is kind of old, didn't realize. thanks for the hint.

is it fine for the master branch?

stof: For now, I'd rather keep the same setup in both branches. It makes it simpler to merge up.

Once the 2.0 rewrite is over and I can stop maintaining 1.x, I will reconsider

stof:
I'm closing this PR:

- the setup with the separate install will be kept for now, as discussed in comments
- as long as we still have the huge baseline for the 1.x implementation, I'd rather keep a fixed version of phpstan so that CI builds don't randomly fail for contributors when phpstan gets updated and some reported errors change. This will also change once we reach the point where we drop the legacy code.

PR #659, opened 2023-09-11
Faster Parser->literal()

staabm:
same as https://github.com/scssphp/scssphp/pull/657, but for 1.x

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -113,6 +113,10 @@ class Parser
      * @var string
      */
     private $buffer;
+    /**
+     * @var string
+     */
+    private $lowercaseBuffer;
     private $utf8;
     /**
      * @var string|null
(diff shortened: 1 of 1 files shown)


src/Parser.php
```diff
@@ -2681,7 +2699,7 @@ protected function value(&$out)
 
         // unicode range with wildcards
         if (
-            $this->literal('U+', 2) &&
+            $this->literal(strtolower('U+'), 2) &&
```
stof, line 2702: why not writing it directly in lowercase ?

staabm: I can do that, if you prefer. 

my reasoning was, that the actual notation in SCSS will have a uppercase U and like that we can easier find it in the codebase

stof:
Do we actually need to lowercase the whole buffer which would double the memory usage of that parser ?

Maybe checking the current character against both `$what[0]` and `strtoupper($what[0])` could be enough to achieve most of the performance gain.

staabm:
seems not worth further investigation

Issue #752, opened 2024-09-13
PHP Version Early Deprecation?


dleffler:
PHP v8.1 isn't deprecated until AFTER Jan 2026, is there a reason the min PHP version was raised to PHP 8.2 a year and a half early?

stof:
The explanation is given in https://github.com/scssphp/scssphp/issues/746

The rule is this one: if someone can make the CI job work on PHP 8.1 for the new compiler (not yet merged, but will soon be), I will happily re-add support for PHP 8.1 (the code is still compatible, and I don't intent to use features requiring PHP 8.2 for now to make this work possible).
However, I will not declare the new version of scssphp compatible with PHP 8.1 if I don't have a CI job to ensure it is indeed the case. And I won't spend my own free time trying to debug why PHP 8.1 segfaults when running the testsuite on GitHub Actions (while it works on my machine).

stof:
Note that this version bump only affects the upcoming version 2.0 of the library.

The 1.x branch still supports PHP 5.6 to 8.3 and this won't change (the code triggers deprecations on PHP 8.4, which cannot be fixed while retaining support for PHP < 7.2, but it works)

TimWolla:
> And I won't spend my own free time trying to debug why PHP 8.1 segfaults when running the testsuite on GitHub Actions (while it works on my machine).

For me running the test suite in a `php:8.1` docker container also segfaulted. Running it in a debug build of PHP 8.1 triggered an assertion failure, a debug build of PHP 8.2 succeeded. I've bisected the change to php/php-src@b3e26c3036a54e9821ea7119c26cdabe484fe36d:

```
b3e26c3036a54e9821ea7119c26cdabe484fe36d is the first new commit
commit b3e26c3036a54e9821ea7119c26cdabe484fe36d
Author: Ilija Tovilo <ilija.tovilo@me.com>
Date:   Fri Apr 19 15:19:02 2024 +0200

    Fix missing handling of CALLABLE_CONVERT in cleanup_unfinished_calls()
    
    Fixes GH-14003

 NEWS                    |  2 ++
 Zend/tests/gh14003.phpt | 24 ++++++++++++++++++++++++
 Zend/zend_execute.c     |  4 ++++
 3 files changed, 30 insertions(+)
 create mode 100644 Zend/tests/gh14003.phpt
```

```
git bisect start
# status: waiting for both good and bad commits
[...]

Issue #746, opened 2024-09-11
[RFC] Considering dropping support for PHP 8.1


stof:
As I'm approaching the end of the work on the dart-sass port, I'm facing an issue in #728: the CI job triggers a PHP segfault when running on PHP 8.1.

I will **not** release version 2 claiming support for PHP 8.1 if I cannot have a CI setup ensuring this is actually the case (no need to try to convince me to change my mind on this, I won't). That segfault seems specific to GitHub Actions apparently (I installed back PHP 8.1 on my machine and was able to run the testsuite, but that's not a viable solution for a CI setup).

I don't have a personal need for the support of PHP 8.1 as all my own projects are already running on PHP 8.2 or 8.3. Because of that, I don't have a personal interest on debugging that segfault before being able to merge #728. I would rather spend time on implement Sass modules after merging the new compiler.

So I'm proposing 2 outcomes:

1. Bumping the min version to PHP 8.2 for the 2.0 version
2. Someone contributing by helping with the CI job on PHP 8.1 so that the CI job in #728 passes

Note that to unblock me, I will likely drop the PHP 8.1 CI job for now (so that I can merge #728 and move forward) while providing a separate draft PR re-enabling to allow for that debugging to happen if someone is interested. However, for at least a few weeks, I will keep writing code based on the PHP 8.1 feature set to make that effort possible.

## Downstream requirements

This analysis is meant to provide additional context to help the decision. However, it does not represent a goal of keeping support for all their minimum versions (we won't go below 8.1 anyway as that decision was already taken).

Top dependents based on https://packagist.org/packages/scssphp/scssphp/dependents?order_by=downloads&requires=require

- Shopware: 8.2
- October CMS: 8.0.2
- Contao: 8.1
- [Typo3 Bootstrap package](https://packagist.org/packages/bk2k/bootstrap-package): 7.4 (but Typo3 now requires PHP 8.2)
- Winter CMS: 8.0.2
- Concrete CMS: 7.3
- [Sass compiler for Typo3](https://packagist.org/packages/wapplersystems/ws-scss): 8.0 (but requires Typo3 12+, which requires PHP 8.1)
- mediawiki/scss: 8.0

Extra known downstream projects:
- [SPIP](https://git.spip.net/spip-contrib-extensions/scssphp): 7.4. The next major version of SPIP requires PHP 8.2
- Gantry 5: 7.1.3
- Grav: 7.3.6 (supported at least through Gantry, maybe by other themes too)
[...]

PR #768, opened 2024-10-19
Fix compatibility with PHP 8.1

TimWolla:
The combination of first-class callables and Exceptions may cause a crash in PHP 8.1 that is fixed for PHP 8.2+ in
php/php-src@b3e26c3036a54e9821ea7119c26cdabe484fe36d.

It specifically can be triggered by the use of `SaasNumber::isComparableTo()` for incomparable numbers and is caused by the
`NumberUtil::fuzzyGreaterThan(...)` FCC as used in `SaasNumber::greaterThan()`.

Fix this issue by calling `$other->coerceValueToMatch($this);` directly, which is the function that is responsible for throwing the `SassScriptException` inside of `SaasNumber::greaterThan()`.

see scssphp/scssphp#746
see scssphp/scssphp#751
see https://github.com/scssphp/scssphp/issues/752#issuecomment-2423857568

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [ '8.2', '8.3' ]
+                php: [ '8.1', '8.2', '8.3' ]
                 extensions: [ '' ]
                 os: [ubuntu-latest]
                 include:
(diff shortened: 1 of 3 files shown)


TimWolla:
> I would prefer using that solution (with a comment explaining that this is done to solve compat with PHP 8.1, potentially with a link to the issue containing the discussion).

Only for `greaterThan()`? For everything in `SassNumber`? Across the whole project?

stof:
Only for cases triggering a segfault

TimWolla:
Adjusted.

stof (approved):

PR #774, opened 2024-12-21
Fix exponentation deprecation for PHP 8.4

stof:
PHP 8.4 has deprecated the usage of the exponentiation operator with 0 as base and a negative exponent, to be consistent with the behavior of `pow` for integers. It also introduces a new `fpow` function to expose the standard behavior for floats (which we want to keep, as Sass numbers are floats).

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -187,6 +187,12 @@ parameters:
 			path: src/Serializer/SerializeVisitor.php
 
 		-
+			message: '#^Strict comparison using \=\=\= between ''\-0'' and ''0'' will always evaluate to false\.$#'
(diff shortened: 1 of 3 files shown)


phpstan-baseline.neon
```diff
@@ -186,6 +186,12 @@ parameters:
 			count: 1
 			path: src/Serializer/SerializeVisitor.php
 
+		-
+			message: '#^Strict comparison using \=\=\= between ''\-0'' and ''0'' will always evaluate to false\.$#'
```
stof, line 190: This is an issue in phpstan itself: https://github.com/phpstan/phpstan/issues/12225

PR #779, opened 2024-12-23
Remove outdated TODO comments

stof:


src/Node/Number.php
```diff
@@ -104,7 +104,6 @@ public function __construct($dimension, $numeratorUnits, array $denominatorUnits
         if (is_string($numeratorUnits)) {
             $numeratorUnits = $numeratorUnits ? [$numeratorUnits] : [];
         } elseif (isset($numeratorUnits['numerator_units'], $numeratorUnits['denominator_units'])) {
-            // TODO get rid of this once `$number[2]` is not used anymore
```
stof, line 107: this TODO has not been resolved. But as the whole class is considered legacy, it will never be done.

Issue #786, opened 2025-01-29
Inline source map lost since release 2.0


ArnaudLigny:
Hello, inline source map don't works anymore since release 2.0.

After a quick search it seems it's related to the following switch: the first case should be `SOURCE_MAP_INLINE` instead on `SOURCE_MAP_NONE`

```diff
switch ($this->sourceMap) {
-    case self::SOURCE_MAP_NONE:
+    case self::SOURCE_MAP_INLINE:
        $sourceMapUrl = 'data:application/json;charset=utf-8,' . Util::encodeURIComponent($sourceMap);
        break;

    case self::SOURCE_MAP_FILE:
        if (isset($this->sourceMapOptions['sourceMapURL'])) {
            $sourceMapUrl = $this->sourceMapOptions['sourceMapURL'];
        }
        break;
}
```

https://github.com/scssphp/scssphp/blob/fe6b678db2fd37f417d8ad01cfca5bc98faf2146/src/Compiler.php#L528C13-L538C14

PR #787, opened 2025-01-30
Fix the rendering of inline source maps

ArnaudLigny:
Fix inline source map (#786).


stof:
good catch. Thanks

I opened another PR to add a test covering the usage of inline source maps, to prevent regressions in the future.

Issue #785, opened 2025-01-26
v2.0 vs v1.13 compile time


skainz92:
I just upgraded from 1.13 to 2.0.

I noticed a significant increase in time that the compiler needs to compile the same file.

With v1.13 i have an average of 116ms
And with v2.0 i have an average of 400ms. 

I upgrade / downgrade via composer. 

Is this maybe a known issue? 
Anybody else experiencing this?

stof:
I have done no effort on performance optimization for the 2.x codebase yet. So this is not unexpected.

The focus of the 2.0.0 release was on correctness, not on performance: the 1.x codebase is [failing 1/3 of the official sass-spec testsuite](https://github.com/scssphp/scssphp/blob/1.x/tests/specs/sass-spec-exclude.txt) excluding module-related tests from it, while 2.0.0 fails only [1 test for a rounding difference](https://github.com/scssphp/scssphp/blob/main/tests/specs/sass-spec-exclude.txt) (the other 2 in the list are related to modules but not detected properly by our logic excluding module-related tests).

Note that I'm not sure how far I would go for performance optimization though. One of our goal is to keep the architecture close enough to the dart-sass architecture, to make it easy to implement new features in a spec-compliant way. Performance optimizations that require changing totally the architecture of the library will likely be rejected because of that.

skainz92:
Thank you for your quick answer!

I came across this issue because i had wrongly disabled caching for the browser, and after upgrading it was immediately clear something with the css files didn't work as before. 
Now with caching enabled, its only on the initial request that takes a little longer - which is ok for me.

Now, i think i will still stay with 1.13, because everything i need works perfectly - and its faster :)

Maybe you could offer one very fast non-compliant version, and one compliant version which is a little bit slower :)
Although maintaining both would be double the work ... 

Either way, thank you very much!
regards,

stof:
I clearly don't want to maintain 2 versions in parallel with no code in common at all (especially given that the 1.x codebase is a mess, meaning that maintaining 1.x is a lot more work than 2.x if maintaining it would mean implementing new features as well).
However, if the 1.x branch works for your use case, you can still use it.

scssphp is indeed not really suited for live rendering of CSS if you don't use caching. But the use cases for compiling Sass to CSS generally benefit easily from caching so this is fine to me.

[...]

PR #800, opened 2025-08-18
[DRAFT] allow parse and compilation caching

Cerdic:
@stof I put this here as a discussion starting point

I implemented a parsing and compilation cache on the v2 in the SPIP plugin, cause it's a needed feature for us, so I made this quick proof of concept, but I'm aware the architecture is not wonderful :)

To be able to implement the parsing cache I added a `Compiler::compileStylesheet` method used by the `Compiler` and `ImportCache` class as a parsing entry point, but this need to make the Compiler class as non final which is not great I guess, and also to pass the Compiler instance to the importCache so it's calling the overloaded method.

Maybe an alternative would be to have `Processor` class with the `compileStylesheet` method. You could provide a overloaded `CacheProcessor` object when instantiating the Compiler, otherwise it would instantiate it's own default `Processor` ?

For the compilation cache, I could work with a decoration of `compileString` and `compileFile` method of the `Compiler` class, except I need some kind of Compilation options signature to differenciate cache. So in my implementation I needed to switch somme properties to `protected` instead of `private`, to allow a `getCompileOptions()` method (see here https://git.spip.net/spip-contrib-extensions/scssphp/-/blob/dev/scssphp2/src/Compiler.php?ref_type=heads#L90) . 
I guess I could also decorate each method changing these properties to keep track of some sort of copy, but maybe you have a better idea?...

The goal is just to have the right entry points in the lib so the caching can be managed from the outside, and we don't need to address this topic in the lib itself. Eventually we could offer later a CompilerCache package that comes on top of the lib for who wants.

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -444,7 +444,8 @@ class Compiler
     /**
      * @throws SassFormatException when parsing fails
      */
-    public static function parseStylesheet(string $contents, Syntax $syntax, ?LoggerInterface $logger = null, ?UriInterface $sourceUrl = null): Stylesheet {
+    public static function parseStylesheet(string $contents, Syntax $syntax, ?LoggerInterface $logger = null, ?UriInterface $sourceUrl = null): Stylesheet
+    {
         return Stylesheet::parse($contents, $syntax, $logger, $sourceUrl);
     }
 


Cerdic:
> The Sass team has an open issue about shared state for such caching of the ImportCache at [sass/sass#2745](https://github.com/sass/sass/issues/2745). Maybe this part should wait until we see what dart-sass does, to avoid diverging implementations. The use case is precisely about multiple entrypoints being compiled in the same process (in webpack builds processing svelte components, in that case).

OK, not sure if it's going to progress fast. What I did is fork the project on my account, and rework a minimal modified version with the ImportParserInterface to make easy to maintain
https://github.com/Cerdic/scssphp/tree/minimal_import_parser_interface
and I'm pointing on this branch in my project https://git.spip.net/spip-contrib-extensions/scssphp/-/blob/dev/scssphp2/composer.json?ref_type=heads

Also the integration is testing the presence of the `setImportParser` method on the Compiler
https://git.spip.net/spip-contrib-extensions/scssphp/-/blob/dev/scssphp2/src/CachedCompiler.php?ref_type=heads#L41
so I can fallback anytime on the official version without breaking the integration (just losing the paring cache in this case)

Also I can remove the branchs on this repo, to keep it clean (and close this PR?)

And maybe at last open an issue about Caching to point this PR and the branchs on my repo, to keep track of the topic and the talks ?

stof:
> Also I can remove the branchs on this repo, to keep it clean (and close this PR?)

yes, that would be great.
Btw, I always use branches in my own fork as the source of PRs, to keep the main repo clean.

> And maybe at last open an issue about Caching to point this PR and the branchs on my repo, to keep track of the topic and the talks ?

you could even open a draft PR from your branch, allowing to hold the discussion about this smaller implementation.

> OK, not sure if it's going to progress fast.

That discussion has stalled because they worked on a simpler project in the meantime (shared resources rather than shared state) which helps improving performances especially for the case of sass-embedded (where they have the overhead of starting the sass compiler process, which can now be reused for multiple compilations).
But it might resume.

In the meantime, I prefer spending time on #785 to optimize some parts of our parser (while staying close to dart-sass).

PR #806, opened 2025-08-21
Sync error messages with dart-sass and enforce them

stof:
This enables assertions on error messages for sass-spec tests (except tests having warnings as well as that's more complex to handle and so will be done separately).
This allowed catching a few cases where I made mistakes when porting the logic from dart-sass.

The 2 excluded tests will be fixed by fixing the support of Unicode content in `scssphp/source-span`, for which I have a partial fix already locally.

### DIFF
--- a/tests/SassSpecTest.php
+++ b/tests/SassSpecTest.php
@@ -205,6 +205,19 @@ class SassSpecTest extends TestCase
         return rtrim($css);
     }
 
+    private static function normalizeErrorOutput(string $output): string
+    {
+        $output = self::normalizeOutput($output);
+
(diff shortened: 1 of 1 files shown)


stof:
After a bunch of attempts, I finally managed to fix the macOS issue (which was indeed related to the fact that the temporary folder is a symlink going inside the `/private/` folder).

For the Windows failure, the official spec runner is also impacted, which I reported at https://github.com/sass/sass-spec/issues/2084. I'll see whether a solution is implemented upstream (which I would then port to our non-official runner). Otherwise, I might look for the proper normalization myself.

PR #815, opened 2025-10-29
Replace `jiripudil/phpstan-sealed-classes` with the phpstan annotation

stof:


stof:
The phpstan failure is due to https://github.com/phpstan/phpstan/issues/13734

PR #823, opened 2025-11-06
Upgrade to league/uri 7.6

stof:
`league/uri` performs 3 changes that are meaningful for us:
- it implements strict validation for the RFC3986 rule requiring a scheme in the base url of the `resolve` algorithm. This makes it match the behavior of the `resolve` method in ext-uri in PHP 8.5. This means that the case 3 described in #819 is no longer provided for us by `league/uri`, which is what this PR needs to implement.
- it deprecates `BaseUri::resolve` (alongside the whole `BaseUri` class) in favor of `Uri::resolve` (and a future `UriInterface::resolve`).
- it exposes some public methods in `UriString` that were previously only private (or maybe protected) in `BaseUri`, which are part of the resolution algorithm. I currently rely on `UriString::removeDotSegments` to implement the new logic for some paths, which is what would make it hard to keep support for `league/uri` 7.5 as is.

TODO:
- [ ] add more test cases for non-RFC3986 cases (and potentially submit them to the dart testsuite as well) to increase test coverage
- [ ] figure out whether the cases calling `UriString::removeDotSegments` directly could be refactored to also be implemented in term of `resolve` with a modified base url followed by some URL manipulation. This would make it possible to keep support for `league/uri` 7.5 (and so to merge this change _before_ the `league/uri` stable release), and would also make that code ready for a (far) future where a new major version of scssphp would use `Uri\Rfc3986\Uri` as its URI representation.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -4,10 +4,6 @@ parameters:
     paths:
         - ./src/
     ignoreErrors:
-        # Ignore error
-        -
-            message: '#^Method ScssPhp\\ScssPhp\\Value\\Sass\w+::assert\w+\(\) has ScssPhp\\ScssPhp\\Exception\\SassScriptException in PHPDoc @throws tag but it''s not thrown\.$#'
-            identifier: throws.unusedType
         # This variable is modified by reference in some Closure and may be assigned null later. Without this type, phpstan does not detect it properly.
         -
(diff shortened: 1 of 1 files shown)


stof:
The TODOs will be done separately as they are not blockers.

PR #841, opened 2026-09-07
Make Base64VLQ MIT-only by reimplementing it from the Source Map v3 spec

tonikruse:
## Summary

`src/SourceMap/Base64VLQ.php` is the only file in this package that is not MIT-licensed. It was ported from Closure Compiler's `Base64VLQ.java` and still carries Google's Apache-2.0 notice. This PR replaces it with an implementation written from the Source Map v3 specification, so that every file in scssphp is covered by the MIT licence declared in `composer.json`.

## Motivation

- A per-file licence notice is binding regardless of the package-level declaration, so today scssphp is effectively "MIT plus one Apache-2.0 file".
- Apache-2.0 is not compatible with GPL-2.0-only. Projects under that licence (a number of PHP shop systems and CMSs among them) cannot ship scssphp as-is, and dependency licence scanners flag the file.
- Making the package uniformly MIT removes that friction for downstream users at no cost to anyone else: the class is `@internal`, has a single static method, and its behaviour is unchanged.

## What changed

- `Base64VLQ::encode()` is reimplemented from the specification. The sign bit and the four lowest magnitude bits form the first Base64 group; the remaining magnitude is split into 5-bit groups; each group is written with the continuation bit as soon as a further group is known to follow.
- Working on the magnitude rather than a sign-folded value keeps every shift non-negative, so the unsigned-shift emulation of the old code is no longer needed and the method terminates for the whole `int` range.
- Public API, namespace, `final`, `@internal` and the call sites in `SingleMapping.php` are untouched. The file header keeps the package's standard MIT block.
- `tests/Base64VLQTest.php` gains vectors for 15, 16, 31, 32 and their negatives, which sit on the four-bit boundary of the first group. The six existing vectors are unchanged.

The previous implementation was of course visible in this repository. The new one was written from the specification and deliberately structured differently (magnitude-first grouping, single pass with look-ahead) rather than by editing the port.

## Behaviour is unchanged

| Check | Result |
|---|---|
| `tests/Base64VLQTest.php` (6 existing + 8 new vectors) | 14/14 pass, and the 8 new vectors also pass against the previous implementation |
| Output compared with the previous implementation for every value in −800 000 … 800 000, 1.2 million random values over the full 64-bit `int` range, and every power-of-two boundary ±1 | 2 800 253 values, 0 differences |
| `2^62`, `−2^62`, `PHP_INT_MAX`, `PHP_INT_MIN` | identical output to the previous implementation |
| Micro-benchmark on typical source-map deltas, PHP 8.4 | ~96 ns/call vs ~112 ns/call before |
| `phpcs` (PSR-12) | clean |
| PHPStan level 8 | no findings in the changed files |


stof:
I fixed the phpstan issue in #842

stof (approved):

PR #843, opened 2026-09-10
Performance improvements

dtdesign:
This is a follow-up of https://github.com/scssphp/scssphp/issues/785. My previous attempts at trying to improve the compile performance did not yield the expected results or were too slim to justify changes. The slow compile times continue to be an annoyance, especially during development where changes would require a recompilation and having to wait an extra couple seconds every time adds up.

I currently have a local test setup where the entire style compilation takes about 4 seconds on PHP 8.4 (Apple M1 Pro) which is annoying to say the least. Applying the changes in this PR cuts this down to close to 2 seconds (~50% reduction) while yielding a byte-for-byte identical stylesheet, verified manually.

In my earlier takes I did notice the lack of memoization as a significant slow down but I tackled the problem at a too low level. Continuing my previous efforts, I threw some company money at Anthropic and involved Fable 5.1 in the process.

**FULL DISCLOSURE**: The commits and the following PR message are created by Fable 5.1. While I did review the changes and had a lengthy conversation with the model, this is in fact an LLM generated change.

## Summary Generated by an LLM

Compiling the stylesheet of WoltLab Suite Core 6.2 with WoltLab Suite Forum 6.2 (~115 imported files, ~4k style rules, 530 variables passed through `Compiler::replaceVariables()` after `ValueConverter::parseValue()`) took ~790 ms with 2.1.0. Profiling showed most of the time went into repeated work rather than into actual compilation. This PR removes the repetition without changing any output; the generated CSS is byte-identical.

| phase | before | after |
|---|---|---|
| `ValueConverter::parseValue()` × 530 | 157 ms | 17 ms |
| `compileString()` | 633 ms | 412 ms |
| total | 790 ms | 429 ms |

### Changes

- **`ValueConverter::parseValue()`** compiled a full stylesheet (`a {b: scssphp-parse-value(($source))}`) per value, constructing a `Compiler`, an `ImportCache` and an `EvaluateVisitor` each time. It now parses `($source)` as a single expression via the new `StylesheetParser::parseExpression()` and evaluates it through the new `EvaluateVisitor::runExpression()`, which mirrors dart-sass' `runExpression()`/`_withFakeStylesheet()`. Error messages and the empty-source → `()` behavior are unchanged.
- **`BuiltInCallable`** re-parsed the argument declarations of the meta functions that `EvaluateVisitor::__construct()` creates for every compilation (~11 parses per compile, ~48 % of `parseValue()`). Parsed declarations are now memoized per signature and module URL.
- **`Selector::__toString()`** serialized the selector on every call; `SimpleSelectorMap::getHash()` alone called it ~43k times per compile. The inspect-mode string is now cached per instance (selectors are immutable and already cache their specificity).
- **`isInvisible()` / `isInvisibleOtherThanBogusCombinators()` / `isBogus()` / `isBogusOtherThanLeadingCombinator()`** allocated a visitor and walked the selector on every call; they run several times for every style rule (extension store, `warnForBogusCombinators()`, serializer). Results are cached on `ComplexSelector` and `SelectorList`, `ComplexSelector::isBogus()` short-circuits on leading combinators, and the list-level visitors go through the cached per-complex methods. `AnySelectorVisitor` uses plain loops instead of closures.
- **Parser**: `consumeIdentifierBody()`, `interpolatedIdentifierBody()` and `whitespaceWithoutComments()` consumed one byte per method call. New `StringScanner::scanCharacterSet()` / `skipCharacterSet()` (`strspn`) consume ASCII runs at once; escapes and non-ASCII bytes still take the original path, so error behavior is unchanged.
- **`ConcreteExtensionStore`**: three `assert()` messages interpolated selectors, which serialized ~11k selectors per compile whenever `zend.assertions=1`; `registerSelector()` did up to three hash lookups per simple selector.
- **Serializer**: strings without newlines, escapes or private-use characters are written in one go; indentation uses `str_repeat()`.

### Verification

- Default test suite passes. `TEST_SASS_SPEC=1` shows the same four pre-existing failures as `main` (escaped `@use`/`@forward`, `random()` deprecation message, float precision).
- PHPStan and phpcs pass.
- Peak memory for the benchmark: 66 MB → 68 MB.


stof:
Can I get more info about the profiling environment ?

- which PHP version ?
- which OPCache configuration (including the JIT config) ?
- which value for `zend.assertions` (a prod environment should have them disabled, i.e. `-1`) ? Our assertions are meant to catch bugs in the compiler and having the selector in the assertion failure makes it easier to debug them.

dtdesign:
That is interesting, I have not thought about the assertions that much. We do enable assertions at runtime in our software because it is a convenient way to enforce things that cannot be statically enforced due to the lack of strict types.

Tests were done in a development environment that by design runs the stock configuration of PHP with the exception of increased memory and POST/upload limits. This allows us to test against what we can expect on the average garbage web hosting environment.

- PHP v8.4.25
- Default OPcache configuration
- `zend.assertions` is enabled by our software.

Cerdic:
If it can help to focus on important things, I did some measurments on compilation time on my real life SCSS styesheet, which is basically a BS5 with some modifications.

* The first commit b424090e4e9cd51d5f123525fa542f281ea64489 gives a clear performance improvment of around 10% compilation time on a total of around 4.5s.
* The second commit 1484149b03fa1df554314e2362759a25ddc6ff0a seems also to give some performance gain but it's less than half of the first commit improvment
* I did not notice real improvment from the third commit b3732ab7568a14b82a89e1f8a1f24b4624735b81

This measurment are made on my local dev computer, PHP 8.4,  `zend.assertions=0`

stof:
I suggest submitting each performance improvement as its own PR, which makes it easier to evaluate them (instead of having to make a all-or-nothing decision). And it would be great to do this split for actual changes rather than for the current commits (where some of those 3 commits actually bundle multiple changes in a single commit).

dragomano:
The improvements are noticeable: compilation time for scssphp/scssphp is about 15% faster (0.4249 → 0.3611 sec), but memory usage increases by about 20% (33.98 → 40.62 MB) as a trade-off. The resulting CSS size stays the same. I have a small benchmark comparing the latest scssphp/scssphp with this fork (WoltLab/scssphp, performance-improvements branch).

With current `scssphp/scssphp`:

| Compiler | Time (sec) | CSS Size (KB) | Memory (MB) |
|------------|-------------|---------------|-------------|
| bugo/scss-php | 0.6502 | 397.68 | 26.79 |
| bugo/scss-php (with cache) | 0.0019 | 397.68 | 26.35 |
| bugo/sass-embedded-php (cli) | 0.1057 | 397.66 | 1.57 |
| bugo/sass-embedded-php | 0.0661 | 397.66 | 2.38 |
| scssphp/scssphp | 0.4249 | 318.38 | 33.98 |
| shyim/sasso-ffi | 0.0176 | 397.66 | 1.07 |

With `https://github.com/WoltLab/scssphp/tree/performance-improvements`:

| Compiler | Time (sec) | CSS Size (KB) | Memory (MB) |
|------------|-------------|---------------|-------------|
| bugo/scss-php | 0.6509 | 397.68 | 27.12 |
| bugo/scss-php (with cache) | 0.0019 | 397.68 | 26.35 |
| bugo/sass-embedded-php (cli) | 0.1035 | 397.66 | 1.57 |
| bugo/sass-embedded-php | 0.0655 | 397.66 | 2.38 |
| scssphp/scssphp | 0.3611 | 318.38 | 40.62 |
| shyim/sasso-ffi | 0.0175 | 397.66 | 1.07 |

dtdesign:
> And it would be great to do this split for actual changes rather than for the current commits (where some of those 3 commits actually bundle multiple changes in a single commit).

The changes are already split up in commits that logically group changes of the same kind. The idea is that while those are touching different places, they follow the same exact intention and principle.

Please don’t get me wrong, I have no issues in splitting this up to make it easier for you to work with. I’m simply not sure how you would like these separated so I get things right the first time.

stof:
> he changes are already split up in commits that logically group changes of the same kind. The idea is that while those are touching different places, they follow the same exact intention and principle.

A commit that is titled `Evaluate parsed values directly instead of compiling a stylesheet` contains 2 unrelated things:

- changing the way parsed values are evaluated for the ValueConverter (what the commit message says)
- adding a static cache for argument declaration parsing in `BuiltinCallable`

I would not call those being the same kind of changes.

The same is true for other commits, where the commit message describes one of the changes (the full commit message mentions other parts, but not the summary which is what we see at first in many places).

And this also makes it harder to evaluate the actual impact of each change. Being faster is good, but if the gain is marginal while making the codebase diverge from the dart-sass code we ported, it will make future updates harder for no actual benefits.

